PR #26, opened 2014-12-29
Add RHEL 7.0

ccaum:
This commit adds a RHEL 7.0 vagrant box provisioner.  Currently, the only RHEL
7 build is nocm box.


stahnma:
@heathseals can you have a look at this?

stahnma:
@ccaum  @heathseals says he's going to add RHEL when he gets to it, and wants it to be the same as the other platforms, so we're going close this in favor of that work.

PR #27, opened 2015-02-20
Refresh Debian Wheezy

heathseals:
This PR refreshes the vagrant boxes for Debian Wheezy.  The OS is bumped from 7.6 to 7.8, both Puppet and PE get version bumps, and a couple issues are addressed with the Virtualbox Guest Additions.


sschneid:
:+1:

PR #30, opened 2015-03-09
(maint) vmware tools vmhgfs patch for Wheezy

heathseals:
inode.c d_alias patch required to successfully compile the vmhgfs module
on Debian Wheezy


sschneid:
:+1:

PR #32, opened 2015-03-19
Solaris 10/11 template tweaks

heathseals:
Still a lot of work to be done with the Solaris templates, but this allows us to build nocm flavors.


stahnma:
LGTM, but I haven't done lots with packer or this workflow.

PR #35, opened 2015-07-13
(RE-4873) initial AIO packaging updates

heathseals:
This PR provides initial AIO/puppet-agent packaging updates to the CentOS, Ubuntu, and Debian PL packer boxes and bumps the version of the PE agent.  Where applicable, the ISO used during the provisioning process has been updated to use a full ISO instead of minimal ( to reduce network usage ).

### DIFF
--- a/manifests/modules/packer/manifests/vmtools.pp
+++ b/manifests/modules/packer/manifests/vmtools.pp
@@ -2,8 +2,8 @@ class packer::vmtools inherits packer::vmtools::params {
 
   if ( $required_packages != undef ) {
     package { $required_packages:
-      ensure     => installed,
-      before     => File[ '/tmp/vmtools' ],
+      ensure => installed,
+      before => File[ '/tmp/vmtools' ],
     }
   }
 
--- a/manifests/modules/packer/manifests/vmtools.pp
+++ b/manifests/modules/packer/manifests/vmtools.pp
@@ -2,8 +2,8 @@ class packer::vmtools inherits packer::vmtools::params {
 
   if ( $required_packages != undef ) {
     package { $required_packages:
-      ensure => installed,
-      before => File[ '/tmp/vmtools' ],
+      ensure     => installed,
+      before     => File[ '/tmp/vmtools' ],
     }
   }
 
--- a/templates/centos-6.6/x86_64.virtualbox.base.json
+++ b/templates/centos-6.6/x86_64.virtualbox.base.json
@@ -4,9 +4,9 @@
     {
       "template_name": "centos-6.6-x86_64",
       "template_os": "RedHat_64",
-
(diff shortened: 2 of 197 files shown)


manifests/modules/packer/manifests/vmtools.pp
```diff
@@ -2,8 +2,8 @@
 
   if ( $required_packages != undef ) {
     package { $required_packages:
-      ensure => installed,
-      before => File[ '/tmp/vmtools' ],
+      ensure     => installed,
+      before     => File[ '/tmp/vmtools' ],
```
ScottGarman: Why the change in  spacing here?

heathseals: they were aligned with another option that ended up not being required.  i'll update shortly.

templates/centos-6.6/x86_64.virtualbox.base.json
```diff
@@ -4,9 +4,9 @@
     {
       "template_name": "centos-6.6-x86_64",
       "template_os": "RedHat_64",
-
-      "iso_url": "http://mirrors.cat.pdx.edu/centos/6.6/isos/x86_64/CentOS-6.6-x86_64-netinstall.iso",
-      "iso_checksum": "2342bc5ea72fc028972aa0f21cee28b4",
+
```
ScottGarman: Should this just be a newline? It looks to me like there are spaces added here.

templates/centos-6.6/x86_64.vmware.base.json
```diff
@@ -4,9 +4,9 @@
     {
       "template_name": "centos-6.6-x86_64",
       "template_os": "rhel6-64",
-
-      "iso_url": "http://mirrors.cat.pdx.edu/centos/6.6/isos/x86_64/CentOS-6.6-x86_64-netinstall.iso",
-      "iso_checksum": "2342bc5ea72fc028972aa0f21cee28b4",
+
```
ScottGarman: Same comment wrt spacing.

ScottGarman:
I'm :+1:  other than the few really minor nit picks about whitespace.

templates/centos-6.6/x86_64.virtualbox.base.json
heathseals: not following.  can you elaborate?

ScottGarman: Is line 7 just a newline, or are there spaces/tabs in it? If it's just a newline, then you can ignore my comment.

heathseals: just a newline.

stahnma:
I can't tell from the diff, but do we apt-get upgrade/yum upgrade after we install from the ISO? This would help for things like heartbleed and POODLEs and such.

heathseals:
@stahnma not consistently across all the distributions.  that's something i noticed during this refresh and have a note to ticket and correct.

stahnma:
Cool.

PR #36, opened 2015-07-15
(RE-4282,4464) Add Fedora Vagrant boxes

heathseals:
This PR adds templates Fedora 21,22 Vagrant boxes.  21 contains nocm and puppet configurations.  22 currently only contains a nocm configuration.

### DIFF
--- a/templates/fedora-21/files/ks.cfg
+++ b/templates/fedora-21/files/ks.cfg
@@ -1,0 +1,36 @@
+install
+cdrom
+lang en_US.UTF-8
+keyboard us
+network --bootproto=dhcp
+rootpw puppet
+firewall --disabled
+authconfig --enableshadow --passalgo=sha512
+selinux --disabled
+timezone UTC
+bootloader --location=mbr
+text
+skipx
+zerombr
+clearpart --all --initlabel
+autopart
+auth  --useshadow  --enablemd5
+firstboot --disabled
+reboot
+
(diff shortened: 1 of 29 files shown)


templates/fedora-21/files/ks.cfg
```diff
@@ -0,0 +1,36 @@
...
+zerombr
+clearpart --all --initlabel
+autopart
+auth  --useshadow  --enablemd5
+firstboot --disabled
+reboot
+
+%packages --ignoremissing
+bzip2
+gcc
```
stahnma: Is there a way we could do what we need to without gcc? Or could we clean it and all it brings in upon completion?

templates/fedora-22/files/ks.cfg
```diff
@@ -0,0 +1,36 @@
...
+zerombr
+clearpart --all --initlabel
+autopart
+auth  --useshadow  --enablemd5
+firstboot --disabled
+reboot
+
+%packages --ignoremissing
+bzip2
+gcc
```
stahnma: Same comment/question as above

templates/fedora-21/files/ks.cfg
heathseals: @stahnma we have to assume gcc is needed for the vmtools.  removing any build tools ( as well as performing an update ) will be part of a future PR.

PR #37, opened 2015-07-22
(RE-5101) add path for puppet-agent binaries

heathseals:
This commit utilizes the profile.d directory to append the path to the
puppet-agent binaries to the user environment.

(maint) fix regression in determining repo type

A module was removed that provided a custom fact regarding yum repo type.
This commit removes the usage of that fact and relies on osfamily
instead.

(maint) bump puppet-agent

puppet-agent 1.2.2 is now available in the public repos.  This commit updates the 
vagrantcloud.yaml to reflect the updated version of puppet.


haus:
:+1:

PR #38, opened 2015-08-07
(maint) nocm box cleanup

heathseals:
PE is utilized for bootstrapping even our nocm box configurations.  The cleanup portion of provisioning should cleanup all the PE bits.   After the last box updates some empty directories were being left behind.   These commits address the cleanup.


mckern:
:+1: Looks like it's not the worst thing ever in the world.

PR #39, opened 2015-08-12
vagrantcloud automation

heathseals:
This PR adds the script that's used to ship and release boxes on VagrantCloud.  This has been managed as a one-off script on the imaging boxes.  Adding three commits to document the lifecycle of the script.


stahnma:
:+1:

PR #41, opened 2015-09-17
Solaris: Set vm_name on base builds

Sharpie:
Packer now defaults to adding a timestamp to the output of builds --- which
makes it impossible for later build stages to guess the name of the artifacts
produced.

This commit fixes the name of Solaris base stage outputs to the format
expected by later build stages.


stahnma:
@heathseals have a look.

stahnma:
This is a good idea. @heathseals is going to have a new PR that is a superset of this for all platforms.

PR #42, opened 2015-09-17
Solaris 11.2: Add release/evaluation package

Sharpie:
Solaris now requires a "licence" package to be installed as a dependency of
many base packages; such as the GCC runtime libraries. Installation of this
package requires an `--accept` flag to be passed to PKG, which breaks
automated builds and installations which aren't expecting a licence check.

This patch adds the `release/evaluation` package to the automated install
manifest for Solaris 11.2.

### DIFF
--- a/templates/solaris-11.2/files/ai.xml
+++ b/templates/solaris-11.2/files/ai.xml
@@ -50,6 +50,7 @@
       -->
       <software_data action="install">
         <name>pkg:/server_install</name>
+        <name>pkg:/release/evaluation</name>
       </software_data>
       <!--
 	    babel_install and slim_install are group packages used to


stahnma:
@heathseals more PRs for you!

shrug:
ping @heathseals

underscorgan:
ping @heathseals

heathseals:
currently working on a number of updates here.  i'll look into this once those are merged.

heathseals:
@Sharpie i'm not running into breakage during the automated installs of 11.2.  where are you seeing the failure?

Sharpie:
@heathseals The failure occurs after the Solaris image is built. When attempting to install Puppet Enterprise, components will be requested which depend on the release/evaluation package. The license check associated with this package will cause PE installation to terminate with an error.

heathseals:
:+1:

PR #43, opened 2015-10-30
Correct path to sftp-server

deployable:
sshd_config currently include's `/usr/lib/openssh/sftp-server` which doesn't exist.


stahnma:
@heathseals ^^

shrug:
@heathseals ^^

heathseals:
@deployable we use this manifest to configure sshd on all puppetlabs-packer templates, so we can't hard code the location of sftp ( since it can vary per distribution ).  the module we use to configure sshd should set this correctly based on distro, but it's obviously not.  i'm looking into this and will have an update soon.

heathseals:
This was fixed in #51 and updated CentOS 7.2 boxes have been published to [Atlas](https://atlas.hashicorp.com/puppetlabs).

PR #45, opened 2015-12-02
(RE-5967) update ec2 configs

heathseals:
These commits update the necessary portions of the packer configs required to produce AMI artifacts for Debian 7.8 and CentOS 7.

### DIFF
--- a/templates/centos-7.0/x86_64.ec2.nocm.json
+++ b/templates/centos-7.0/x86_64.ec2.nocm.json
@@ -2,13 +2,15 @@
   "variables": {
     "aws_access_key": "",
     "aws_secret_key": "",
-    "pe_url": "'https://pm.puppetlabs.com/cgi-bin/download.cgi?ver=latest&dist=el&arch=x86_64&rel=7'",
+    "pe_url": "{{env `PE_URL`}}",
     "provisioner": "ec2",
-    "source_ami": "ami-c7d092f7",
-    "ami_description": "puppetlabs-centos-7.0-x86_64 {{timestamp}}",
-    "ami_name": "puppetlabs-centos-7.0-x86_64 {{timestamp}}",
+    "source_ami": "ami-d440a6e7",
+    "ami_description": "puppetlabs-centos-7.1-x86_64 {{timestamp}}",
+    "ami_name": "puppetlabs-centos-7.1-x86_64 {{timestamp}}",
     "required_modules": "puppetlabs-stdlib saz-ssh",
-    "user_data": "#cloud-config\ndisable_root: 0\n"
+    "user_data": "I2Nsb3VkLWNvbmZpZwpkaXNhYmxlX3Jvb3Q6IDAK",
+    "_comment": "#cloud-config\ndisable_root: 0\n"
+
   },
 
   "builders": [
@@ -25,6 +27,7 @@
       "source_ami": "{{user `source_ami`}}",
       "region": "us-west-2",
       "ssh_timeout": "10000s",
+      "ssh_pty": "true",
       "user_data": "{{user `user_data`}}",
       "run_tags": {
         "created_by": "puppetlabs-packer",
@@ -42,7 +45,7 @@
   "provisioners": [
     {
       "type": "shell",
-      "execute_command": "chmod +x {{ .Path }}; {{ .Vars }} {{ .Path }} {{user `required_modules`}}",
+      "execute_command": "chmod +x {{ .Path }} && {{ .Vars }} {{ .Path }} {{user `required_modules`}}",
       "environment_vars": [
         "PE_URL={{user `pe_url`}}"
       ],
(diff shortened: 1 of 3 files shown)


templates/centos-7.0/x86_64.ec2.nocm.json
```diff
@@ -2,13 +2,13 @@
...
     "provisioner": "ec2",
-    "source_ami": "ami-c7d092f7",
-    "ami_description": "puppetlabs-centos-7.0-x86_64 {{timestamp}}",
-    "ami_name": "puppetlabs-centos-7.0-x86_64 {{timestamp}}",
+    "source_ami": "ami-d440a6e7",
+    "ami_description": "puppetlabs-centos-7.1-x86_64 {{timestamp}}",
+    "ami_name": "puppetlabs-centos-7.1-x86_64 {{timestamp}}",
     "required_modules": "puppetlabs-stdlib saz-ssh",
-    "user_data": "#cloud-config\ndisable_root: 0\n"
+    "user_data": "I2Nsb3VkLWNvbmZpZ1xuZGlzYWJsZV9yb290OiAwXG4"
```
kbarber: What is this Heath? Does this change the disable_root status? We had that there so that we could log in as root from our tests, otherwise it did that weird ubuntu thing where "everything is sudo no root login" from the get-go :-). I'm just wary this won't work with our test framework as it is, thats all, not without changes.

templates/centos-7.0/x86_64.ec2.nocm.json
```diff
@@ -42,7 +43,7 @@
   "provisioners": [
     {
       "type": "shell",
-      "execute_command": "chmod +x {{ .Path }}; {{ .Vars }} {{ .Path }} {{user `required_modules`}}",
+      "execute_command": "sudo mkdir -p /tmp/packer-puppet-masterless/manifests/modules && sudo chown -R centos /tmp/packer-puppet-masterless && sudo chmod +x {{ .Path }} && {{ .Vars }} sudo -E -S bash {{ .Path }} {{user `required_modules`}}",
```
kbarber: Yeah, this kind of complication is what I mean :-). But I'm worried about beaker also.

templates/centos-7.0/x86_64.ec2.nocm.json
heathseals: @kbarber it's the same user_data, but base64 encoded.  see mitchellh/packer#2616.

stahnma:
:+1:

templates/centos-7.0/x86_64.ec2.nocm.json
kbarber: Hmm thats interesting. Perhaps the disable_root setting isn't being passed through as we think in your new base64 encoded form, as this should do it on boot I think. I wonder if the 'fixing' of this root access is done through your packer pp & shell scripts rather than via this setting? That would explain why it works properly later on. Just feels like we're missing something here, I recall doing this to stop that sudo proliferation on purpose.

heathseals: @kbarber it's possible.  i'll take a look user_data.

kbarber: Huh:

```
2.0.0-p647 :006 > Base64.encode64("#cloud-config\ndisable_root: 0\n")
 => "I2Nsb3VkLWNvbmZpZwpkaXNhYmxlX3Jvb3Q6IDAK\n"
```

That's different.

kbarber: Might want to add a comment somewhere, that provides the original string also, so we can flip it back when 0.9.0 is released and let packer do the work. I think "_comment" fields are a usual way of commenting JSON, not sure if packer will puke on extra items though :-).

heathseals:
```
17:32 $ echo -n '#cloud-config\ndisable_root: 0\n' | openssl base64
I2Nsb3VkLWNvbmZpZ1xuZGlzYWJsZV9yb290OiAwXG4=
```
@kbarber I had dropped the padding.  I'll update the PR and strip out the sudo, that'll certainly simplify things.  thanks for questioning that. : )

kbarber: No worries, cut my teeth on this stuff for a few hours I seem to recall :-). I did kind of find it awesome that the same API worked for both Redhat & Debian boxes though ...

heathseals:
@kbarber updated.  i also had an issue with my encoding due to new lines.  that's all sorted, root logins work, and sudo has been removed from the packer templates.

kbarber:
@heathseals this visibly looks okay, did you want me to test an image with our beaker tests or something?

kbarber:
@heathseals if it helps as a final test, if you have AMI's ready to test, we can put them into our relevant beaker config files, and test them out for realz, as we have PR testing: https://github.com/puppetlabs/puppetdb/blob/stable/config/image_templates/ec2.yaml

Either raise a PR for us, or give me the AMI id's and I'll create a test PR for you. This would make a great smoke test on these images I think, and allow you to correct this early before merging in. Then, we can also merge in our PR and use these AMI's immediately. What do you think? Should be pretty easy in theory.

kbarber:
PR for puppetdb here, PR tests just started: https://github.com/puppetlabs/puppetdb/pull/1768

kbarber:
@heathseals tests pass on the redhat box fwiw: http://jenkins-enterprise.delivery.puppetlabs.net/job/enterprise_puppetdb_init-multijob_githubpr-stable/76/

I just realised we don't also test on debian7 atm, due to capacity melarky. But we do on merge.

Anyway, I think this is close enough, if you give us final AMI's we'll test it fully.

+1 on this patch, although debian 7 is still an unknown for now.

PR #48, opened 2015-12-08
(RE-6109) Add packer configs for Debian 8

heathseals:
This commit adds Debian 8.2/jessie nocm and puppet-agent
packer configs for both VMware and VirtualBox.


shermdog:
:-1:

PR #49, opened 2015-12-11
(RE-6104) packer::vsphere support for CentOS and Fedora

heathseals:
This PR adds initial support for shipping CentOS and Fedora artifacts using the vSphere post-processor.  It's primarily intended to be used with the [vmpooler](https://github.com/puppetlabs/vmpooler).

- [x] Fixups/improvements to the packer::vsphere module
  - [x] update to support new bootstrap method ( using vmtools )
  - [x] add bootstrap templates and ssh authorized_keys
  - [x] yum repo management
- [x] Add vSphere Configs
   - [x] CentOS 5
   - [x] CentOS 6
   - [x] CentOS 7
   - [x] Fedora 21
   - [x] Fedora 22

### DIFF
--- a/manifests/modules/packer/manifests/vsphere/params.pp
+++ b/manifests/modules/packer/manifests/vsphere/params.pp
@@ -1,6 +1,7 @@
 class packer::vsphere::params {
 
-  $repo_mirror    = 'http://osmirror.delivery.puppetlabs.net'
+  $repo_mirror = 'http://osmirror.delivery.puppetlabs.net'
+  $loweros     = downcase($::operatingsystem)
 
   case $::operatingsystem {
     'Ubuntu': {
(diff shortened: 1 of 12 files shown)


shermdog:
I think you may want to consider adding the interface restart to the NetworkManager portion of the bootstrap.  https://github.com/puppetlabs/pooler-template-bootstrap/blob/master/centos-7/vcloud-bootstrap.rb#L25-L29

I observed in later versions of NetworkManager that restarting it does not always force it to renew the DHCP lease and as such the hostname does not get back into DNS, leading to clone failures with vmpooler.

heathseals:
i wasn't able to reproduce the DHCP/DNS hostname issue with these templates, but since it has been an issue recently i don't see any reason not to force another interface restart.

heathseals:
@shermdog updated.

stahnma:
:+1:

PR #50, opened 2015-12-17
(RE-6139) add Debian 8 vsphere/vmpooler configs

heathseals:
This commit adds vsphere-nocm packer templates.  The vSphere
artifact is meant to be used with the vmpooler.


sschneid:
:+1:

PR #51, opened 2015-12-18
(CPR-233) Add packer configs for CentOS 7.2

heathseals:


stahnma:
:+1: assuming it works.

PR #60, opened 2016-03-04
(maint) build pipeline updates

heathseals:
This PR is mainly intended to add support for ensuring certain system packages are the latest available, without bumping the minor version of a distribution ( I really dislike this and will revisit at some point ).  We also bump the PE agent-only install to 3.8.4, enable the future parser, and update arguments passed to the VMware Tools install ( needed for the move to Fusion 8 in our build pipeline ).

### DIFF
--- a/manifests/base.pp
+++ b/manifests/base.pp
@@ -1,3 +1,4 @@
+include packer::updates
 include packer::sshd
 include packer::networking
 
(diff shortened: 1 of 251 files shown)


stahnma:
:+1: I feel like there could be less copy/pasta code in there, but I am not sure how that would work.

PR #61, opened 2016-03-10
(RE-6532) vangrant box updates

heathseals:
These commits bump the version of most of our publicly hosted Debian, CentOS, and Ubuntu vagrant boxes on Atlas.  The boxes get an updated version of puppet, an updated version of the PE 3.8 agent-only install, as well as OS updates to address recent security vulnerabilities.  The automation used to release boxes on Atlas has also been updated -- vmware_fusion should not be specified as a provider and vmware_desktop should be used instead.

### DIFF
--- a/bin/vagrantclouder
+++ b/bin/vagrantclouder
@@ -115,13 +115,11 @@ $config['arches'].each do |arch|
             config['s3'] = 'nocm'
         end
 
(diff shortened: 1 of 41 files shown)


shermdog:
:+1: 

@melissa you want to give this a look and hit the button?

PR #63, opened 2016-03-11
(maint) trusty fixups

heathseals:
trusty was unintentionally bumped a point release during the previous
box update commits.  this reverts back to the original point release
and updates the preseed to prevent it being upgraded to the latest point.


underscorgan:
:+1:

PR #64, opened 2016-03-17
(RE-6745) add Scientific 7.2 vsphere/vmpooler config

heathseals:
This commit adds a Scientific 7.2 vsphere-nocm packer template.  The
vSphere artifact is meant to be used with the vmpooler.


stahnma:
:+1:

PR #65, opened 2016-03-18
(maint) bump the xenial daily build sha

heathseals:
This should address some issues were seeing with dhcpc on xenial.  Also update the module we're using for vsphere networking fixups ( so it's consistent on both debian and redhat based operating systems ).

### DIFF
--- a/templates/ubuntu-16.04/i386.vmware.base.json
+++ b/templates/ubuntu-16.04/i386.vmware.base.json
@@ -3,10 +3,10 @@
   "variables":
     {
       "template_name": "ubuntu-16.04-i386",
-      "template_os": "ubuntu-64",
+      "template_os": "ubuntu",
 
-      "iso_url": "http://releases.ubuntu.com/16.04/ubuntu-16.04-server-i386.iso",
-      "iso_checksum": "fa97768bdc3f198b82180d39bf0c26f021ab716f5da98094cd220771095e3394",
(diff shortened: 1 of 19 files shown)


shermdog:
:+1:

templates/ubuntu-16.04/i386.vmware.base.json
```diff
@@ -3,10 +3,10 @@
   "variables":
     {
       "template_name": "ubuntu-16.04-i386",
-      "template_os": "ubuntu-64",
+      "template_os": "ubuntu",
 
-      "iso_url": "http://releases.ubuntu.com/16.04/ubuntu-16.04-server-i386.iso",
-      "iso_checksum": "fa97768bdc3f198b82180d39bf0c26f021ab716f5da98094cd220771095e3394",
+      "iso_url": "http://cdimage.ubuntu.com/ubuntu-server/daily/current/xenial-server-i386.iso",
+      "iso_checksum": "c3892e5e3f634ef42c454bfbfb20fb891d2c2215a738582828aed84d9c67a7d1",
```
ScottGarman: If this url points to the current daily image, does that mean the checksum is going to change daily and then break something?

shermdog: @scottgarman correct.  Until they publish beta 2 these SHAs are only good for a couple of days.  Still good to track.

PR #66, opened 2016-03-21
(maint) vsphere el7 fixups

heathseals:
This commit disables the firewall for vsphere/vmpooler el7 artifacts and
extends the networking fixups to all of el7.


stahnma:
:+1:

PR #77, opened 2016-06-17
(RE-7330) Update Fedora 24 Gold RC + Use AIO to provision

shermdog:
The time has come to use the AIO vs older 3.8 binary.  This is necessary as default providers and library support has changed in modern operating systems, such as DNF and Python3
(Note we'll update this soonish to remove NFS all together and just do the fun stuff locally - better for the community and our sanity).

Added headless as monitoring the build over VNC is great for troubleshooting.

Set target vsphere NIC to vmxnet3 and removed dependency on specifying interface in bootstrap

### DIFF
--- a/manifests/modules/packer/manifests/vsphere.pp
+++ b/manifests/modules/packer/manifests/vsphere.pp
@@ -9,6 +9,23 @@ class packer::vsphere inherits packer::vsphere::params {
     password => "$qa_root_passwd"
   }
 
+  case $::osfamily {
+    redhat: {
+      if $::operatingsystemrelease in ['24', '25'] {
+        Package {
+          provider => 'dnf',
+        }
+
+        file { '/etc/dhclient.conf':
+          owner   => 'root',
+          group   => 'root',
+          mode    => '0644',
+          source  => 'puppet:///modules/packer/vsphere/dhclient.conf',
+        }
+      }
+    }
+  }
+
   package { $ruby_package:
     ensure => present,
   }
--- a/scripts/bootstrap-aio.sh
+++ b/scripts/bootstrap-aio.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+if [[ ${TEMPLATE} == "fedora-22"* ]]; then
+  echo "Updating rpcbind..."
+  dnf -y upgrade rpcbind
+  systemctl enable rpcbind.socket
+  systemctl restart rpcbind.service
+fi
+
+# Create mount point and required directories
+mkdir -p /opt/puppetlabs
+mkdir /var/cache/puppet
+
+if [ -n "${PUPPET_NFS}" ]; then
+  # Mount NFS share if PUPPET_NFS set
+  echo "Mounting AIO via NFS..."
+  mount -o ro -t nfs "${PUPPET_NFS}/${TEMPLATE}/opt/puppetlabs" /opt/puppetlabs
+else
+  echo "The environment variables PUPPET_NFS must be provided to provision AIO." >&2
+  exit 1
+fi
+
+# Show Puppet version
+printf 'Puppet ' ; /opt/puppetlabs/puppet/bin/puppet --version
+
+# Installed required modules
+for i in "$@"
+do
+  /opt/puppetlabs/puppet/bin/puppet module install $i --modulepath=/tmp/packer-puppet-masterless/manifests/modules --vardir='/var/cache/puppet' >/dev/null 2>&1
+done
+
+printf 'Modules installed in ' ; /opt/puppetlabs/puppet/bin/puppet module list --modulepath=/tmp/packer-puppet-masterless/manifests/modules
--- a/scripts/bootstrap-aio.sh
+++ b/scripts/bootstrap-aio.sh
@@ -1,22 +1,44 @@
 #!/bin/bash
 
-if [[ ${TEMPLATE} == "fedora-22"* ]]; then
-  echo "Updating rpcbind..."
-  dnf -y upgrade rpcbind
-  systemctl enable rpcbind.socket
-  systemctl restart rpcbind.service
-fi
+tmp_dir="/tmp/packer-puppet-masterless"
+mkdir -p "$tmp_dir"
+# Download and install puppet-agent
 
-# Create mount point and required directories
-mkdir -p /opt/puppetlabs
-mkdir /var/cache/puppet
-
-if [ -n "${PUPPET_NFS}" ]; then
-  # Mount NFS share if PUPPET_NFS set
-  echo "Mounting AIO via NFS..."
-  mount -o ro -t nfs "${PUPPET_NFS}/${TEMPLATE}/opt/puppetlabs" /opt/puppetlabs
+if [ -n "${PC_REPO}" ]; then
+  echo "Downloading and installing AIO from repo..."
+  # Determin if this is RPM or DEB and Do the Right Thing (tm)
+  if [[ ${PC_REPO} == *".rpm"* ]] ; then
+    curl --silent "${PC_REPO}" -o $tmp_dir/pc-repo.rpm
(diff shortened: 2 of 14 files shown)


witjoh:
@shermdog @stahnma 
Could this be merged ? I'm trying to add lvagrant-libvirt to the templates, and I would love to use the AIO scripts. See also https://tickets.puppetlabs.com/browse/CPR-371
thx

manifests/modules/packer/manifests/vsphere.pp
```diff
@@ -9,6 +9,23 @@
     password => "$qa_root_passwd"
   }
 
+  case $::osfamily {
+    redhat: {
+      if $::operatingsystemrelease in ['24', '25'] {
+        Package {
+          provider => 'dnf',
+        }
```
heathseals: Do we need to specify the provider type here?

scripts/bootstrap-puppet.sh
```diff
@@ -36,7 +36,8 @@ do
   /opt/puppet/bin/puppet module install $i --modulepath=/tmp/packer-puppet-masterless/manifests/modules >/dev/null 2>&1
 done
 
-if [[ ${TEMPLATE} == "fedora-23"* ]]; then
+# Any supported Fedora now needs this
+if [[ ${TEMPLATE} == "fedora-2"* ]]; then
```
heathseals: I think only fedora-23 needs the dnf shim ( until we move it to AIO ).

heathseals:
@shermdog instead of leaving it on disk, i've been cleaning up the old network interface config script and replacing it with a correct copy [here](https://github.com/puppetlabs/puppetlabs-packer/blob/master/manifests/modules/packer/manifests/vsphere/networking.pp).  Should we do that here as well?  purge ens33 and setup ens160?

scripts/bootstrap-puppet.sh
shermdog: This is left over from Mike's original PR from the pre-AIO work - i didn't feel like changing it :)

heathseals:
@shermdog keeps things simple, but it leaves us with a ifcfg-ens33 config file which doesn't correspond to a valid device once it's shipped to vSphere with a vmxet3 device.  it's more about keeping things tidy.

manifests/modules/packer/manifests/vsphere.pp
heathseals: Ahh.  I remember now.  I wasn't setting this on my builds.  I assume it was just falling back to yum.

shermdog:
@heathseals I personally rather have an extraneous file laying around in exchange for having a less fragile provisioning process in the future.

shermdog:
Closing this one for now so that I can work on some changes without GitHub spamming folks with notifications :)

stahnma:
@heathseals please look over. We think it looks good.

witjoh:
Minor detail : in scripts/cleanup-aio.sh you could also remove the downloaded rpm/deb.

templates/fedora-24/i386.vsphere.nocm.json
```diff
@@ -0,0 +1,107 @@
...
+      "name": "{{user `template_name`}}-{{user `provisioner`}}-vsphere-nocm",
+      "vm_name": "packer-{{build_name}}",
+      "type": "vmware-vmx",
+      "headless": true,
+      "source_path": "output-{{user `template_name`}}-{{user `provisioner`}}/packer-{{user `template_name`}}-{{user `provisioner`}}.vmx",
+      "ssh_username": "root",
+      "ssh_password": "puppet",
+      "ssh_port": 22,
+      "ssh_wait_timeout": "10000s",
+      "shutdown_command": "/sbin/halt -h -p",
```
heathseals: on more recent vSphere artifacts, i've been adding an annotation so we have a better idea of how/when it was built :
```
"vmx_data": {
        "annotation": "{{user `template_name`}}-{{user `version`}} built {{isotime}}"
},
```

templates/fedora-24/x86_64.vsphere.nocm.json
```diff
@@ -0,0 +1,107 @@
...
+    {
+      "name": "{{user `template_name`}}-{{user `provisioner`}}-vsphere-nocm",
+      "vm_name": "packer-{{build_name}}",
+      "type": "vmware-vmx",
+      "headless": true,
+      "source_path": "output-{{user `template_name`}}-{{user `provisioner`}}/packer-{{user `template_name`}}-{{user `provisioner`}}.vmx",
+      "ssh_username": "root",
+      "ssh_password": "puppet",
+      "ssh_port": 22,
+      "ssh_wait_timeout": "10000s",
```
heathseals: same comment as above regarding an annotation 
```
"vmx_data": {
        "annotation": "{{user `template_name`}}-{{user `version`}} built {{isotime}}"
},
```

templates/fedora-24/x86_64.vsphere.nocm.json
```diff
@@ -0,0 +1,107 @@
...
+      "shutdown_command": "/sbin/halt -h -p",
```
heathseals: same comment as above regarding an annotation 
```
"vmx_data": {
        "annotation": "{{user `template_name`}}-{{user `version`}} built {{isotime}}"
},
```

scripts/cleanup-aio.sh
```diff
@@ -0,0 +1,18 @@
...
+  rpm -e puppet-agent
+elif [[ ${PUPPET_AIO} == *".deb"* ]] ; then
+  dpkg -P puppet-agent
+else
+  echo "Unsupported AIO package format" >&2
+  exit 1
+fi
+# Remove any Puppet-related files and directories
+rm -rf /etc/puppetlabs
+rm -rf /opt/puppetlabs
```
heathseals: also clean up the puppet-agent.{rpm,deb}?

scripts/bootstrap-aio.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+if [[ ${TEMPLATE} == "fedora-22"* ]]; then
+  echo "Updating rpcbind..."
+  dnf -y upgrade rpcbind
+  systemctl enable rpcbind.socket
+  systemctl restart rpcbind.service
+fi
```
heathseals: we can get rid of the rpcbind update for f22 since we're deprecating PUPPET_NFS.

scripts/bootstrap-aio.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Download and install puppet-agent
+
+if [ -n "${PC_REPO}" ]; then
+  echo "Downloading and installing AIO from repo..."
+  # Determin if this is RPM or DEB and Do the Right Thing (tm)
+  if [[ ${PC_REPO} == *".rpm"* ]] ; then
+    wget --quiet "${PC_REPO}" -O pc-repo.rpm
+    rpm -Uhv pc-repo.rpm
+    dnf install -y puppet-agent
```
heathseals: should we add some logic to utilize `yum` if `dnf` isn't available?

heathseals:
i've commented on a few things, but overall i like what i see here.

scripts/cleanup-aio.sh
shermdog: This is now cleaned up by nature of the tmp_dir being cleaned up by the cleanup-packer script.  This is intentional as we can just omit the cleanup-aio for boxes we want to ship with Puppet pre-installed.

templates/fedora-24/i386.vsphere.nocm.json
shermdog: Done.

templates/fedora-24/x86_64.vsphere.nocm.json
shermdog: Done.

PR #78, opened 2016-06-23
(RE-7448) Add VMWare Tools to build sequence

jcoconnor:
This installs VM Build Tools as part of the base installation so that
VMXNET3 device drivers are available for later use.

This is a first pass - i.e. only 50% reliable, but happy for comments,
suggestions until I have a more robust method.


jcoconnor:
I have just pushed a change up with the following mods:

1. Mount VMWare tools in the boot by using additional ```vmx_data```
2. Installing VMWare tools in the unattended install.
3. Specifying vmxnet3 as the interface, as once VMWare tools is installed, the network comes up.

I was able to locate the ```windows.iso``` under ```/Applications/VMware Fusion.app/Contents//Library/isoimages/windows.iso```

Comments on this approach please @McdonaldSeanp @glennsarti @heathseals

templates/windows-2012r2/x86_64.vmare.base.json
```diff
@@ -67,7 +64,12 @@
         "time.synchronize.shrink": "1",
         "time.synchronize.tools.startup": "1",
         "time.synchronize.tools.enable": "1",
-        "time.synchronize.resume.host": "1"
+        "time.synchronize.resume.host": "1",
+        "sata0.present": "TRUE",
+        "sata0:0.present": "TRUE",
+        "sata0:0.autodetect": "TRUE",
+        "sata0:0.deviceType": "cdrom-image",
+        "sata0:0.fileName": "/Applications/VMware Fusion.app/Contents//Library/isoimages/windows.iso"
```
glennsarti: I would probably use a template variable instead of a hard coded path here.  That way it could be overridden at build time.

glennsarti:
After review, this is substantial process for something that is generally lightweight.  A couple of issues;

* Currently assumes that the packer build is on a Mac Laptop with Fusion.  As we intend this to be automated it probably won't be on the same platform, so this needs to be more platform agnostic

* Ideally you want to install VMWare Tools as late as possible.  The version of VMWare tools you need is mainly determined by the Hypervisor version e.g. Fusion has different Tools than ESX 5.0 so It would be best to keep that installation out of base and in standard

* The required files for the VMX3 nic is about 140KB, which is very small so to be honest, we could just add them to this git repository and add them to the A:\ for automated installation.  No need to install Tools at all.

If adding them to the repo is not an option, would we consider a pre-packer phase where we pull down assets/artefacts from sources prior to running packer?

jcoconnor:
@glennsarti A few points:

1. The automated platform also uses Mac with Fusion so for the moment we are ok there - I will certainly use a variable here for the ISO definition as you suggest - but there may be a better way.
2. The Fusion version of vmware tools works fine with ESX - we have been using it for a while now.
3. A pre-packer phase is always an option too -  but generally all of the other builds are build-able from the packer command line without pre-processing, so would like to keep it that way.

John

glennsarti:
Ok..LGTM.. can always discuss it further later.

templates/windows-2012r2/x86_64.vmare.base.json
jcoconnor: Done

jcoconnor:
@glennsarti As per our discussion, I have made a few changes in sequencing and account handling including:
1. box-starter is executed in the generalize-packer unattended install only.
2. Auto-Login accounts moved to the specialize phase.
3. Execute generalise-packer at the end of the first unattended install to do the sys prep.
4. Add in the packer-shutdown for the final shutdown (as we don't want to run the sysprep again).

This appears to run from start to finish, but would appreciate your comments if you get a chance to test it this evening.

Thanks - John

heathseals:
the base image seems to build fine for me.  failed the first time due to fusion not having yet pulled down the windows tools iso, other than that no issues.  not sure how best to validate it actually did what it was suppose to do, but i was able to login using the puppet account and ensured the tools were running.

templates/windows-2012r2/files/autounattend.xml
```diff
@@ -1,326 +1,186 @@
...
+                    </CreatePartitions>
+                    <ModifyPartitions>
+                        <!-- EFI system partition (ESP) -->
+                        <ModifyPartition wcm:action="add">
+                            <Order>1</Order>
+                            <PartitionID>1</PartitionID>
+                            <Label>System</Label>
+                            <Format>FAT32</Format>
+                        </ModifyPartition>
+                        <!-- MSR partition does not need to be modified -->
```
McdonaldSeanp: Is this comment out of date or does the below not actually modify the MSR?

templates/windows-2012r2/files/autounattend.xml
```diff
@@ -1,326 +1,186 @@
...
+        </component>
+        <component name="Microsoft-Windows-Security-SPP-UX" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS" xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+            <SkipAutoActivation>true</SkipAutoActivation>
+        </component>
+    </settings>
+    <settings pass="oobeSystem">
+        <component name="Microsoft-Windows-Shell-Setup" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS" xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+          <UserAccounts>
+            <AdministratorPassword>
+                <Value>Qu@lity!</Value>
```
McdonaldSeanp: Are there security concerns leaving this value hardcoded in a public repo?

templates/windows-2012r2/files/autounattend.xml
```diff
@@ -1,326 +1,186 @@
...
+                        <PartitionID>3</PartitionID>
+                    </InstallTo>
+                </OSImage>
+            </ImageInstall>
+            <UserData>
+                <!-- Product Key from http://technet.microsoft.com/en-us/library/jj612867.aspx -->
+                <ProductKey>
+                    <!-- Do not uncomment the Key element if you are using trial ISOs -->
+                    <!-- KMS Setup keys  https://technet.microsoft.com/en-us/library/jj612867.aspx -->
+                    <Key>D2N9P-3P6X9-2R39C-7RTCD-MDVJX</Key>
```
McdonaldSeanp: Same question as below, is there concerns leaving this key in a public repo?

McdonaldSeanp:
Will run a test build today. The build scripts themselves look good minus the security concerns I brought up above. If those are fine I am 👍

templates/windows-2012r2/files/autounattend.xml
jcoconnor: This one is more problematic, as we may need to seed the build with a key.
I will see if there is a better way of hiding this.

templates/windows-2012r2/files/autounattend.xml
jcoconnor: Had intended to obfuscate this using the Image Management tool (todo)

jcoconnor: Obfuscated.

templates/windows-2012r2/files/autounattend.xml
jcoconnor: Done

jcoconnor:
@McdonaldSeanp & @glennsarti Have done some minor fixes as suggested by Sean.
Have also squashed commits down to 3 (which roughly follow the 3 tickets being submitted).

Still need to remove the Windows key - watch this space.

templates/windows-2012r2/files/autounattend.xml
glennsarti: This is NOT a private key.  This is a well known public install key for KMS enabled operating system

Ref
https://technet.microsoft.com/en-us/library/jj612867(v=ws.11).aspx

McdonaldSeanp:
Things look good. I feel this is good to merge 👍

glennsarti:
Good to merge for me.  👍

PR #80, opened 2016-06-29
Fix typo in Debian preseed partitioning config option

DanHam:
Spotted and hopefully fixed a typo in the Debian preseed config files


shermdog:
:+1: @heathseals ?

PR #81, opened 2016-07-07
(RE-7385) Structure for configuring host

jcoconnor:
Initial structure to establish environment to configure host using
puppet-agent (masterless).

Adding:
1. install-win-packages.ps1 - to install all the windows packages
2. clean-host.ps1 - Host cleanup script
3. Rename top level Puppet manifest & json files to match naming
   conventions.
4. Use C:\Packer staging directory structure for Downloads, Puppet
   modules & manifests and logs (this will be deleted in a later ticket
   once the overal packer process is completed.
5. Edits to get puppet-agent structure operational. Later tickets will
   look at the actual configuration details.

Any TODO's left in script files are for addressing in later tickets.


jcoconnor:
@craig-gomes @McdonaldSeanp This is a gist of the current run status (with latest push up): https://gist.github.com/jcoconnor/42f16381fafdcfb499741a1ad6bfc9bf

jcoconnor:
@glennsarti Thanks for pointers today. I have revised the layout of the modules directory and puppet appears to be running a bit better now. Gist at: https://gist.github.com/jcoconnor/36014002435421a6c95ee2bfb253f0ab

Still shutdown problems at end, but will try and sort that out on Monday.

/cc @McdonaldSeanp

jcoconnor:
@McdonaldSeanp @glennsarti @heathseals This is ready for merge pending any review comments.
It appears to reliably clone and update the resultant machine using the puppet manifest and modules.
Plenty still to do in later tickets.

jcoconnor:
Gist of build: https://gist.github.com/jcoconnor/f10e1c43e0a5024d069733721d834683

To test this you need you need to have a pre-built base windows template.

So use the following commands
```
packer build x86_64.vmware.base.json
packer build x86_64.vmware.vsphere.cygwin.json
```

scripts/windows/cleanup-host.ps1
```diff
@@ -0,0 +1,25 @@
+# Cleanup tasks to ready this machine for production.
+
+$ErrorActionPreference = 'Stop'
+
+. A:\windows-env.ps1
+
+
+Write-Host "Uninstalling Puppet Agent..."
+Start-Process -Wait "msiexec" -ArgumentList "/i C:\Packer\Downloads\puppet-agent.msi /qn /norestart PUPPET_AGENT_STARTUP_MODE=manual"
```
glennsarti: Shouldn't that be `/x` instead of `/i` ?

jcoconnor: Oops - yes - it should be - fix coming up now.

scripts/windows/windows-env.ps1
```diff
@@ -0,0 +1,16 @@
+
+$ErrorActionPreference = 'Stop'
+
+# Ensure TEMP directory is set and exists. Git.install can fail otherwise.
```
glennsarti: Does this happen in Packer?

jcoconnor: TEMP directory seems to exist ok, so will remove this.
Main reason for env script is to define common variables used across scripts.
(e.g. Packer Location) - at the moment its a placeholder, but I will move to use it more in later tickets.

jcoconnor: Done.

scripts/windows/cleanup-host.ps1
jcoconnor: Done

scripts/windows/windows-env.ps1
glennsarti: I'm not sure those `$ENV` calls survive Start-Process calls anyway.  They;re being put into the process space I believe.  May need more testing if they're required later.

scripts/windows/shutdown-packer.bat
```diff
@@ -1,7 +1,7 @@
 @ECHO OFF
 SETLOCAL
 
-SET LOGFILE=%SYSTEMROOT%\TEMP\shutdown-packer.bat.log
+SET LOGFILE=C:\Packer\Logs\shutdown-packer.bat.log
```
glennsarti: Can't say I'm a fan of hardcoded drive paths, but it should be ok in this instance.

jcoconnor: Thats going to go in the Env (and may end up reverting it back to TEMP anyway).

jcoconnor:
@glennsarti & @McdonaldSeanp Have just pushed up change to VMX_Data to use scsi device instead of sata (IDE) device as per: https://github.com/mitchellh/packer/issues/1242

This seems to fix issue. Have done this as a maint commit as it is a change to the base build rather than the vsphere one.

heathseals:
@jcoconnor a few suggestions, mainly related to some variables and the output path we use elsewhere.

within the **base build**, consider setting the variables :  ```template_name``` to match the distribution/arch being built : ```"template_name": "windows-2012r2-x86_64"``` and also set ```"provisioner": "vmware"```.  within the builder, set ```"name": "{{user `template_name`}}-{{user `provisioner`}}"```.

for **downstream builds**, also set the ```template_name``` and ```provisioner``` variables.  within the builder section, set ```name``` to include the provisioner and config type : ```"name": "{{user `template_name`}}-{{user `provisioner`}}-vsphere-cygwin"``` and also modify the path to the source vmx : ```"source_path": "output-{{user `template_name`}}-{{user `provisioner`}}/packer-{{user `template_name`}}-{{user `provisioner`}}.vmx"```.

this will fit into some assumptions we already make within the build tooling.  and speaking of tooling, it'd be great to automate the creation of the sections of packer configs that should be the same across templates.

glennsarti:
@heathseals made some good points.  Would we consider doing that in this PR or raise a ticket of work for later?  Given we're trying to get something working first

jcoconnor:
Will look at it tomorrow morning as I think it does fall into remit of this
ticket. I alternative is to ticket for next sprint. In any case it needs to
be done. What do you think?

jcoconnor:
@heathseals I have updated the naming as you suggested above - am running it through build at the moment.

heathseals:
👍

PR #84, opened 2016-07-13
(RE-7460) Puppet cfg fixes for misc configuration

jcoconnor:
Set of fixes to original puppet code by Glenn for miscellaneous registry
and other configuration settings to match WIndows Template
Configuration.

The fixes are:
1. Typo in Windows Error Reporting
2. Add Power/Hibernation configuration settings
3. Created C:\Crash_Dump directory

Have also addressed Administrator Autologin (RE-7512)

### DIFF
--- a/manifests/windows/windows_template/manifests/local_group_policies.pp
+++ b/manifests/windows/windows_template/manifests/local_group_policies.pp
@@ -5,7 +5,7 @@ class windows_template::local_group_policies ()
 
     windows_group_policy::gpupdate { 'GPUpdate':
     }
-    
+
     windows_group_policy::local::machine { 'DisableServerManagerAtLogon2012':
         key   => 'Software\Policies\Microsoft\Windows\Server\ServerManager',
         value => 'DoNotOpenAtLogon',
@@ -37,7 +37,7 @@ class windows_template::local_group_policies ()
     }
 
     windows_group_policy::local::machine { 'DisableWER':
-        key   => 'Software\Policies\Microsoft\Windows Error Reporting',
+        key   => 'Software\Policies\Microsoft\Windows\Windows Error Reporting',
         value => 'Disabled',
         data  => 1,
         type  => 'REG_DWORD',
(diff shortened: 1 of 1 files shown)


manifests/windows/windows_group_policy/manifests/local/machine.pp
```diff
@@ -4,7 +4,7 @@
   $value     = '',
   $type      = 'REG_SZ',
   $data      = '',
-  $logoutput = false,
+  $logoutput = true,
```
jcoconnor: This will be reverted once I have finished debugging.

manifests/windows/windows_template/manifests/local_group_policies.pp
```diff
@@ -49,23 +49,62 @@
...
+    registry::value { 'AdministratorDefaultUserName':
+        key   => 'HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Winlogon',
+        value => 'DefaultUserName',
+        data  => 'Administrator',
+        type  => 'string'
+    }
+    registry::value { 'AdministratorDefaultPassword':
+        key   => 'HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Winlogon',
+        value => 'DefaultPassword',
+        data  => 'Qu@lity!',
```
McdonaldSeanp: Same as we did in the other PR, should this password be hidden?

McdonaldSeanp:
Other than the one password question, This looks good to me 👍

manifests/windows/windows_template/manifests/local_group_policies.pp
jcoconnor: Probably - wonder if I should roll this change out for the moment or hide it through environment variable (tho this is really insecure).

In any case, this approach won't work for Windows-10, so might be better to see if I can just do this in the unattended installation (which does allow you to obfuscate the password).

jcoconnor:
@McdonaldSeanp I have backed out the password change - its in a separate commit - will deal with that separately.

McdonaldSeanp:
Perfect, 👍

glennsarti:
👍 for merge

PR #85, opened 2016-07-13
(RE-7566) Add Ubuntu 10.04 for internal use

shermdog:
We still maintain Ubuntu 10.04 pools internally.  Move these over to packer as part of the disk increase epic.


jcoconnor:
@shermdog The .json files look ok - can't comment on the ubuntu bits.

stahnma:
👎 closing

PR #88, opened 2016-07-22
(WIP) (RE-7458) Install Chrome, Notepad and Sysinternals

jcoconnor:
Use chocolatey to install packages including Sysinternals packages [RE-7457](https://tickets.puppetlabs.com/browse/RE-7457)
At the moment, the chocolatey.org repo is used.
We may move this to using our local nexus repository in the future (will be handled in separate ticket).

PR #89, opened 2016-07-26
(RE-7663) Don't use boxstarter for later packer configuration

jcoconnor:
Boxstarter (Restart) does not play well with packer once winrm is
enabled, so change the second phase configuration to work without
boxstarter.

This means adding some additional restarts into the packer provisioning
sequence.

This also incorporates #88 which will be closed.

Tickets addressed:

1. [RE-7663](https://tickets.puppetlabs.com/browse/RE-7663) - Fix Boxstarter restart issue
2. [RE-7458](https://tickets.puppetlabs.com/browse/RE-7458) - Install Chrome, Notepad and other utils
3. [RE-7457](https://tickets.puppetlabs.com/browse/RE-7457) - Install and license Sysinternals.

### DIFF
--- a/scripts/windows/cleanup-host.ps1
+++ b/scripts/windows/cleanup-host.ps1
@@ -6,18 +6,21 @@ $ErrorActionPreference = 'Stop'
 
 
 Write-Host "Uninstalling Puppet Agent..."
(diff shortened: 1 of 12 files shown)


glennsarti:
:+1:  for merge for me

McdonaldSeanp:
LGTM

PR #98, opened 2016-08-05
(RE-7829) Create Windows 2016 tech. prev. 5 base template 

McdonaldSeanp:
This is pretty identical to the 2012r2 work.

NOTE: this will change the location of generalize-packer.autounattend.xml so it can be platform specific between 2012 and 2016


jcoconnor:
Trying it now

jcoconnor:
👍 
This builds nice and cleanly. Good work.

PR #97, opened 2016-08-04
(RE-7590) Install and Configure Cygwin

jcoconnor:
Install Cygwin 2.4.0 from a locally downloaded repository.
See RE-7600 for background on why we don't want to use a later version of Cygwin.
This was obtained off fruitbat.org and saved to buildsources.

The Cygwin SSH configuration is done along with the registry and service configuration and appropriate updates are made to the post-cloning procedures.

PR #99, opened 2016-08-09
(RE-7671) Cleanup Machine for deployment

jcoconnor:
1. Clear WinSxS (update) space in base image.
2. Use sdelete to zero out unused space.
3. Remove Chocolatey (RE-7671)
4. Delete Temp and other unecc files.

Note - the sdelete is only done following the win-update cleanup as this
is the single largest action and gives about a 6G reduction of disk
space. Cleanmgr was evaluated, but only did minor cleanup (<1G so wasn't
considered worth the effort of using for Windows-2012).

The DISM commands used here only work for Windows 2012+, so an
alternative cleanup is required for earlier windows versions.

This PR is rebased on top of #97 to pick up ```ARCH``` definitions needed for it and supersedes it, so it also includes:

Install Cygwin 2.4.0 from a locally downloaded repository.
See RE-7600 for background on why we don't want to use a later version of Cygwin.
This was obtained off fruitbat.org and saved to buildsources.

The Cygwin SSH configuration is done along with the registry and service configuration and appropriate updates are made to the post-cloning procedures.


jcoconnor:
This is a WIP in PR for the final cleanup tasks.
Comments appreciated /cc @McdonaldSeanp @glennsarti @cowofevil

scripts/windows/cleanup-host.ps1
```diff
@@ -12,17 +12,55 @@ Start-Process -Wait "msiexec" -ArgumentList "/x $PackerDownloads\puppet-agent.ms
...
 $pageFileMemoryKey = "HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Memory Management"
 Set-ItemProperty -Path $pageFileMemoryKey -Name PagingFiles -Value ""
 
-
 # Run Clean disk.
+@(
+    "$env:localappdata\Nuget",
+    "$env:localappdata\temp\*",
+    "$env:windir\logs",
+    "$env:windir\panther",
```
glennsarti: I would not recommend kill panther.  I've seen issues where if this folder is missing subsequent SYSPREPs fail

scripts/windows/cleanup-host.ps1
```diff
@@ -12,17 +12,55 @@ Start-Process -Wait "msiexec" -ArgumentList "/x $PackerDownloads\puppet-agent.ms
...
+        if(Test-Path $_) {
+            Write-Host "Removing $_"
+            Takeown /d Y /R /f $_
+            Icacls $_ /GRANT:r administrators:F /T /c /q  2>&1 | Out-Null
+            Remove-Item $_ -Recurse -Force -ErrorAction SilentlyContinue | Out-Null
+        }
+    }
+
+Write-Host "defragging..."
+Optimize-Volume -DriveLetter C
```
glennsarti: I'm am not a fan of defragging virtual disks.  Due to the various IO redirections, having contiguous blocks inside the Guest is mostly useless.

scripts/windows/cleanup-host.ps1
```diff
@@ -12,17 +12,55 @@ Start-Process -Wait "msiexec" -ArgumentList "/x $PackerDownloads\puppet-agent.ms
...
+
+Write-Host "0ing out empty space..."
+& sdelete64.exe  -z c:
```
glennsarti: Does sdelete require EULA?

May want to add OS Arch detection so you use 64bit and 32bit versions appropriately

scripts/windows/cleanup-host.ps1
jcoconnor: Noted and changed.

scripts/windows/cleanup-host.ps1
jcoconnor: It also fails - have removed it.

scripts/windows/cleanup-host.ps1
jcoconnor: ```-accepteula``` was in fact added, but not pushed up to the repo.
Will add OS detection once I have rebased this on top of #97 which has ARCH detection.

I have also moved ```delete``` to immediately following the WinSxS update as this single operation clears the most amount of space (>6GB). It also takes significant time, so is worth pushing back in the process.

geoffnichols:
LGTM

jcoconnor:
Minor update pushed up to clean the [WER directories](http://serverfault.com/questions/21777/archived-and-queued-windows-error-reporting) - this frees up about 1G of data.

McdonaldSeanp:
👍

PR #104, opened 2016-08-16
(RE-8755) Copy Cygwin setup to system32 directory

jcoconnor:
This is address a beaker permissions issue.

Need to copy the Cygwin_setup across to C:\Windows\system32 to overcome an issue with the curl command in beaker testing. The beaker tests do need fixing, but this is an interim workaround to prevent pipeline breakages.

PR #106, opened 2016-08-17
(RE-7855) Copy Cygwin setup to system32 directory

jcoconnor:
This is address a beaker permissions issue.

Need to copy the Cygwin_setup across to C:\Windows\system32 to overcome
an issue with the curl command in beaker testing. The beaker tests do
need fixing, but this is an interim workaround to prevent pipeline
breakages.

NB - this replaces #104 which was titled incorrectly.


heathseals:
👍

PR #108, opened 2016-08-19
Clean windows update leftovers / only use one virtual disk file

McdonaldSeanp:
These commits will clean out files leftover from windows updates, and update the build scripts to only create one virtual disk file.


mckern:
👍 This seems sane, since you can't leave comments in a JSON blob.

PR #113, opened 2016-08-26
(RE-7993) Change Disk Handling

jcoconnor:
A series of changes to the disk handling - the main one is to reduce the amount of disc traversed by ```sdelete``` to reduce the overall build time. 

(RE-7993) Change Disk Handling
1. Change disk type to Sparse Monolothic (Type 0), https://www.vmware.com/pdf/VirtualDiskManager.pdf
2. Move DISM cleanup to winrm section.
3. Limit initial disk size to 20G (to limit sdelete)
4. Extend partition to full size at final system preparation.

### DIFF
--- a/scripts/windows/cleanup-host.ps1
+++ b/scripts/windows/cleanup-host.ps1
@@ -68,6 +68,16 @@ wevtutil clear-log Security
 wevtutil clear-log Setup
 wevtutil clear-log System
 
+# TODO run sdelete a final time - only a suggestion as it may be useful to pare out the
+# extra space released by the delete commands above.
+
+# Extend C: partition to full extend - this is predicated on the
+$size = (Get-PartitionSupportedSize -DriveLetter C)
+$sizemax = $size.SizeMax
+Write-Host "Setting Drive C partition size to $sizemax"
+Resize-Partition -DriveLetter C -Size $sizemax
+
+
 # Sleep to let console log catch up (and get captured by packer)
 Start-Sleep -Seconds 20
 #End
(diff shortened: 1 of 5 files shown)


scripts/windows/cleanup-host.ps1
```diff
@@ -68,6 +68,16 @@ wevtutil clear-log Security
 wevtutil clear-log Setup
 wevtutil clear-log System
 
+# TODO run sdelete a final time - only a suggestion as it may be useful to pare out the
+# extra space released by the delete commands above.
+
+# Extend C: partition to full extend - this is predicated on the
+$size = (Get-PartitionSupportedSize -DriveLetter C)
```
glennsarti: I won't let this stop this PR but `Get-PartitionSupportedSize` only exists in Powershell on Windows 2012 and above.  For Windows 2008 and Windows 7 VMs you'll need to use alternate methods. (WMI and diskpart)

templates/windows-2012r2/x86_64.vmware.base.json
```diff
@@ -34,13 +34,15 @@
       "shutdown_timeout": "1h",
       "guest_os_type": "windows8srv-64",
       "disk_size": 61440,
+      "disk_type_id": "0",
```
glennsarti: For the sake of documentation.  Disk Type 0 is `sparse monolithic`, default is 1 `sparse split`.

https://www.vmware.com/pdf/VirtualDiskManager.pdf

glennsarti:
:+1: for merge

scripts/windows/cleanup-host.ps1
jcoconnor: @glennsarti Tx for the pointer.
I'm really targeting this at Win-2012+ for the moment.
There are few things that need to be re-engineered for Win-2008 etc - (e.g. you need to use ```cleanmgr``` instead of ```dism```) so I have a ticket open for that when I get round to those platforms.

PR #115, opened 2016-09-01
(RE-7926) Add Windows 2016 RTM Build

jcoconnor:
This is a copy of the TP5 build pointing to the latest RTM image.
The only major change was to make sure the KMS keys were specified
correctly as available in:
https://technet.microsoft.com/en-us/library/jj612867.aspx


McdonaldSeanp:
LGTM, it's just a copy of tp5 so :+1:

jcoconnor:
@McdonaldSeanp @glennsarti Testing this on jenkins at the moment.
May need to work on the timeout for the boot command.
But suggest we merge this first and tune later.

PR #116, opened 2016-09-01
(RE-7926) Set Install Image correctly

jcoconnor:
Set the Windows Server Edition in the Unattend.xml so that we don't have
to use buggy boot-prompts.

This is an amendment to #115


jcoconnor:
@McdonaldSeanp & @glennsarti This is an amendment to #115 - it sets the Server Edition in the Unattended Answer file rather than in menu selection on the initial boot.

PR #117, opened 2016-09-02
(RE-7957) Change Powershell module to 1.0.6

jcoconnor:
This needs associated changes to the puppet-configure script.
The configure scripts changes are useful - the need to revert to PS
1.0.6 is under review.

Need to tidy up version handling using PS splatting but putting this out for review for the moment.
/cc @McdonaldSeanp @glennsarti


McdonaldSeanp:
:+1:

PR #118, opened 2016-09-06
(RE-7834) macOS Sierra (10.12) packer template

heathseals:
This PR contains the initial work for macOS Sierra packer templates.  It currently produces two artifacts -- a base VM and a vSphere artifact for use with the vmpooler.  

This heavily utilizes Tim Sutton's [osx-vm-templates](https://github.com/timsutton/osx-vm-templates) for both the preparation of the bootable ISO as well as scripts used by the shell provisioner.  Some modifications have been made to prepare the vSphere artifact for use with the vmpooler


stahnma:
👍

PR #126, opened 2016-09-13
(MAINT) Fixup Casing in init directory name

jcoconnor:
To fix Init and init directories - move all files to "init" and fixup json references in packer scripts.


jcoconnor:
@heathseals Could you give this a quick overview and merge as a trivial request.
Its an issue that @glennsarti picked up last night where we have two "init" directories with different casings.

PR #131, opened 2016-09-15
(RE-8013) Update to use new signing key

melissa:
Soon we will be changing out the key used to sign the new public repos.
This code will break once that happens without this update, but this
update should be merged as close as possible to when the repos are
switched over to the new key


mckern:
Trivial merge invoked

PR #132, opened 2016-09-26
(IMAGES-48) Packer Windows-2012 Automation

jcoconnor:
Initial load of Win-2012 files for base win-2012 build.
This is mostly a copy of the win-2012r2 template with the following main changes:

1. DISM command for 2012 doesn't support /resetbase (leads to larger image unfortunately).
2. Initial image size needs to be 30GB

Remainder of template code appears to operate satisfactorily.

### DIFF
--- a/scripts/windows/clean-disk-dism.ps1
+++ b/scripts/windows/clean-disk-dism.ps1
@@ -2,9 +2,17 @@ $ErrorActionPreference = 'Stop'
 
 . A:\windows-env.ps1
 
-# Cleanup Windows Update area after all that (may need reboot)
+$WindowsVersion = (Get-WmiObject win32_operatingsystem).version
+# Cleanup Windows Update area after all that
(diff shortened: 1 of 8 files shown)


jcoconnor:
This builds without issue using VMWare on Mac/Desktop.
Running it through a Jenkins build (which won't complete until tomorrow).

McdonaldSeanp:
Nothing looks out of place 👍

jcoconnor:
@glennsarti & @McdonaldSeanp The only significant difference I can see compared to win-2012 is that the DISM command doesn't support /ResetBase - this means that the generated image size if somewhat larger than that for Win-2012R2 :(

jcoconnor:
@glennsarti & @McdonaldSeanp I have been able to deploy this template to the Dev Pooler without any issue, so this is ready for merge.

PR #135, opened 2016-10-04
(MAINT) Save Win-2012 slipstream commands

jcoconnor:
The Win-2012 slipstream was produced manually - this is a save of the
main batch/powershell commands that were used to create it.
This will be used as a basis to script a better solution for future
slipstream work.


glennsarti:
This is a good start.

glennsarti:
This is not the final version either.  It's a place holder.

jcoconnor:
Post submit comment.
Tx @branan and @glennsarti 
I previously discussed with with @heathseals as he has some similar files he wants to submit and we came up with the ```util``` directory for this purpose.

PR #136, opened 2016-10-06
(IMAGES-76) Change Post-Clone to sysprep operation

jcoconnor:
The Run Once key in the registry is armed to execute a sysprep command
which runs the sysprep and then the post-clone configuration script.

Add unattended file for all current OS along with an update to the
vsphere json file to upload the unattended file

Removed HKCU registry key configuration to since Default User keys will
now be used to create the Administrator Profile after the sysprep.
Needed to add one registry fix to the post-clone-config to correct the
background screen setting.

### DIFF
--- a/scripts/windows/windows-env.ps1
+++ b/scripts/windows/windows-env.ps1
@@ -61,7 +61,6 @@ param (
 # As noted elsewhere, the intention to to replace all Powershell registry calls with Puppet code
 
 Function Set-UserKey($key,$valuename,$reg_type,$data) {
-  Write-Host "Setting user registry entry: $key\$valuename"
-  reg.exe ADD "HKCU\$key" /v "$valuename" /t $reg_type /d $data /f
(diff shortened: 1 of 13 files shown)


scripts/windows/windows-env.ps1
```diff
@@ -62,6 +62,6 @@ param (
 
 Function Set-UserKey($key,$valuename,$reg_type,$data) {
   Write-Host "Setting user registry entry: $key\$valuename"
-  reg.exe ADD "HKCU\$key" /v "$valuename" /t $reg_type /d $data /f
+# reg.exe ADD "HKCU\$key" /v "$valuename" /t $reg_type /d $data /f
```
glennsarti: Probably should just remove this

jcoconnor: Done - and also updated ```Write-Host``` command

PR #139, opened 2016-10-26
(IMAGES-367) Switch to Win-2016 GA Release

jcoconnor:
The GA Release ISO is available so switching to use this.
Also renaming directory to remove "rtm" from name as its no longer appropriate

Note - need to slipstream in the first CU update as this appears to
hang the build if updated normally (KB31949798)


ScottGarman:
:+1:

PR #144, opened 2016-11-04
(IMAGES-369) Special template for slipstream

jcoconnor:
Specialised template that:
1. Boots from base Windows ISO
2. Downloads and installs all windows updates from a WSUS Server.
3. Applies the updates to an offline copy of the windows image.
4. Generates a new ISO Slipstream image from the offline updated copy.

This image can then be exported to ```osmirror``` to be used for the Packer/Windows build process.

It is intended for manual use on a desktop at the moment.

### DIFF
--- a/scripts/windows/generate-slipstream.ps1
+++ b/scripts/windows/generate-slipstream.ps1
@@ -1,0 +1,93 @@
+# Generates Slipstream ISO from image
+param (
+  [string]$OSName = "UNKNOWN",
+  [string]$ImageIndex = "1"
+)
+
+$ErrorActionPreference = 'Stop'
+
+. C:\Packer\Dism\windows-env.ps1
+
+New-Item -ItemType directory -Force -Path C:\Packer\Dism\$OSName
+
+$UpdateDirectory = "$ENV:Windir\SoftwareDistribution\Download"
+$DismBase = "$PackerStaging\Dism"
+$MountPoint = "$DismBase\Mount"
+$WinIsoPath = "$DismBase\$OSName-SlipStream.iso"
+$WinDistPath = "$DismBase\$OSName"
+$WinImageFile = "$WinDistPath\sources\install.wim"
+
+# Install ADK
+Write-Host "Install Win ADK"
+Download-File http://buildsources.delivery.puppetlabs.net/windows/winadk/adksetup_win2012r2.exe  $PackerDownloads\adksetup_win2012r2.exe
+Start-Process -Wait "$PackerDownloads\adksetup_win2012r2.exe" -ArgumentList "/quiet /norestart /features OptionId.DeploymentTools"
+Write-Host "Win ADK Installed"
+
+# Need extra disk for this bit.
+$IsoGen = "C:\Program Files (x86)\Windows Kits\8.1\Assessment and Deployment Kit\Deployment Tools\amd64\Oscdimg\oscdimg.exe"
+if ($psversiontable.psversion.major -gt 2) {
+  $size = (Get-PartitionSupportedSize -DriveLetter C)
+  $sizemax = $size.SizeMax
+  Write-Host "Setting Drive C partition size to $sizemax"
+  Resize-Partition -DriveLetter C -Size $sizemax
+}
+else {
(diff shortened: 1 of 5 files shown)


jcoconnor:
Ping @puppetlabs/agent-team This PR is ready for review.

It appears to work nicely for Windows 2012R2 - I have generated a Slipstream ISO from it which is pre-loaded with 192 windows updates.

templates/windows-2012r2/x86_64.vmware.base.json
```diff
@@ -3,9 +3,9 @@
     "template_name": "windows-2012r2-x86_64",
 
     "provisioner": "vmware",
-    "iso_url": "http://osmirror.delivery.puppetlabs.net/iso/windows/en_windows_server_2012_r2_with_update_x64_dvd_6052708.iso",
```
glennsarti: The changes to this file should probably be in a separate commit.  So we can tell the difference between creating a slipstreamed ISO and consuming it.

jcoconnor: Done

glennsarti:
I'm not sure on the name  `x86_64.vmware.slipstream.json`.  Packer can use mutliple infra builders so putting this in the name seems pointless.  Also the slipstreamed output is not a VMware dependency.  I would've thought the name `x86_64.slipstream.json` would be more appropriate.

jcoconnor:
@glennsarti Can we leave it as is for the moment. I think you are correct that we should come up with a better naming scheme, so perhaps ticket it ?

glennsarti:
Ok.

templates/windows-2012r2/files/windows-2012r2-x86_64-vmware.slipstream.package.ps1
```diff
@@ -0,0 +1,76 @@
+$ErrorActionPreference = "Stop"
```
glennsarti: We could probably move this the `\scripts` directory and drop the `windows-vmware` prefix.  This boxstarter package should be able to handle any OS.

scripts/windows/generate-slipstream.ps1
```diff
@@ -0,0 +1,93 @@
...
+		  Write-Host "Update $Cab Succeeded"
+    } else {
+		  Write-Host "***** Update $Cab FAILED *******"
+      $Cab.Name | Out-File -FilePath $DismBase\Updates-Failed.log -Append
+    }
+}
+
+dism /unmount-wim /mountdir:$MountPoint /commit
+
+& $IsoGen -m -o -u2 -udfver102 -bootdata:"2#p0,e,b$WinDistPath\boot\etfsboot.com#pEF,e,b$WinDistPath\efi\microsoft\boot\efisys.bin" $WinDistPath $WinIsoPath
```
glennsarti: I assume you manually retreive the ISO file after the build is finished?

glennsarti:
Raised IMAGES-373 to review naming standards.

jcoconnor:
Yes.
Didn't want to out any credentials or certs in the repo.
If you have any ideas on a good way to transfer it out let me know.

Will add a comment in the code that it needs to be manually retrieved.

glennsarti:
We could probably pass them through as template user params and then in the powershell call, add them as optional parameters with a default of no copy.

glennsarti:
Could also look at the File provisioner;  https://www.packer.io/docs/provisioners/file.html

Note the direction property so you can move files from remote to local.

jcoconnor:
@glennsarti I missed the direction property in the file upload.
However, the file transfer performance using WinRM for packer is very very poor, so I think the best way to handle this would be to use scp or similar (or even a good old file share). 

Suggest we ticket it for a future improvement.

templates/windows-2012r2/files/windows-2012r2-x86_64-vmware.slipstream.package.ps1
jcoconnor: Done

glennsarti:
All queries have been answered and additional work has been ticketted. LGTM

PR #148, opened 2016-11-18
(IMAGES-336) Create Japanese template

jcoconnor:
Add build for Japanese Windows-2012R2.
This is deliberately using the Copy/Pasta approach, i.e. making a full
copy of the template files for English-Win-2012R2.

A further ticket will look to combining the templates once the
complexities of a localised build have been evaluated.

### DIFF
--- a/templates/windows-2012r2-ja/README.md
+++ b/templates/windows-2012r2-ja/README.md
@@ -1,0 +1,16 @@
+# puppetlabs-packer
+
+### About
+
+This is the Windows 2012R2 Packer Template
+
+### VM settings
+
+
+## Documentation
+
+The Confluence Documentation for the process is at [Windows/Packer Imaging Process](https://confluence.puppetlabs.com/display/QE/Packer+Generation+of+Windows+Templates+for+VMPooler)
+
+### Issues
+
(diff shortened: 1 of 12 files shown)


jcoconnor:
Have verified that this builds and deploys on ```vmpooler-dev```
Cygwin also appears to initialise correctly.

McdonaldSeanp:
@jcoconnor do we have any way to verify that the copy-paste approach didn't do anything strange ( i.e. can we verify that this approach produces what would be a normal japanese installation? ). Is there any way to reproduce manually a "normal" install of a machine for japanese?

/cc @glennsarti @Iristyle

jcoconnor:
@McdonaldSeanp If you mean the initial image creation - the only changes I made were to change the local to ja-JP instead of en-US (with en-US as the fallback locale) - otherwise its a standard ```AutoUnattended.xml``` file.

We could create a windows-Japanese installation manually by booting off the ISO directly, but not sure how we would do a valid comparison.

McdonaldSeanp:
@jcoconnor it sounds like there aren't significant differences in the installations for english and jp (such as the names for program Files, programdata and such). So I think that the fact that the update process works  ( which you mentioned to me on hipchat ) is a reasonable indicator that the installation is good

McdonaldSeanp:
👍 

Testing seems to indicate this is GTG

PR #154, opened 2016-11-29
(IMAGES-382) Initial commits for Fedora 25

stahnma:


templates/fedora-25/i386.vsphere.nocm.json
```diff
@@ -0,0 +1,110 @@
...
+        "PUPPET_AIO={{user `puppet_aio`}}"
+      ],
+      "scripts": [
+        "../../scripts/bootstrap-aio.sh"
+      ]
+    },
+
+    {
+      "type": "file",
+      "source": "../../hiera",
```
heathseals: apparently packer no longer likes if a source directory is missing.  we should probably just create this directory and leave it empty.

templates/fedora-25/x86_64.vsphere.nocm.json
```diff
@@ -0,0 +1,110 @@
+{
+
+  "variables":
+    {
+      "template_name": "fedora-25-x86_64",
+      "version": "0.0.3",
+
+      "provisioner": "vmware",
+      "required_modules": "puppetlabs-stdlib example42-network",
+      "puppet_aio": "https://yum.puppetlabs.com/fedora/f24/PC1/i386/puppet-agent-1.8.1-1.fedoraf24.i386.rpm",
```
heathseals: this should point to the x86_64 package

stahnma:
I completed base builds of both architectures locally, probably should have said that.

PR #155, opened 2016-11-30
(maint) Ensure virtualbox builds are headless

stahnma:


ScottGarman:
Should the commit msg specify the 16.10 builds are headless? :+1:  otherwise

PR #164, opened 2016-12-14
(maint) rename vsphere templates

heathseals:
Our linux vsphere templates do not match our typical
arch.builder.post-processor.config naming.  This commit renames those
templates to match our standard.


jcoconnor:
👍

PR #168, opened 2016-12-27
(CPR-395) Update kernel package with osfamily RedHat

vinzent:
the kernel package is required to be updated at install time so it matches the kernel-devel version, required for VBoxGuestAddtions installation


stahnma:
@vinzent thanks.

PR #169, opened 2016-12-27
(CPR-394) Add Fedora 24 and 25 with VirtualBox packer templates

vinzent:


stahnma:
@heathseals opinions here?

vinzent:
@stahnma @heathseals I came to the conclusion that I better use distro upstream images (fedora/25-cloud-base, centos/7, ...) instead of puppet provided boxes. 

Puppet boxes are really outdated (centos 6.6 instead of 6.9, no fedora at all, centos 7.2 instead of centos 7.3) and it seems nobody cares about them (https://groups.google.com/forum/#!topic/puppet-dev/B3bFZCmxarA). 

I've switched to the distro provided images without problems using https://github.com/puppetlabs/beaker-puppet_install_helper to install puppet.

stahnma:
@vinzent thanks, and I understand. We basically do these when we get a chance, which isn't all the time. If you need current stuff, going with upstream is probably better nearly always.

PR #171, opened 2017-01-11
(maint) update travis to use trusty

heathseals:
Update travis to use the trusty container rather than the default of
precise.  This will address an issue with the old curl in precise and
newly issued hashicorp certs.

Related to https://github.com/travis-ci/travis-ci/issues/7131


jcoconnor:
👍 Merging to re-enable Travis

PR #179, opened 2017-02-08
(IMAGES-39) Add Windows-2008 Template

jcoconnor:
Create the base and vSphere templates for Windows-2008

This packer build differs a bit from the other windows:
1. Slipstream had to be prepared manually in a few stages due to the
   necessity of adding Powershell, .Net 3.5 SP1 and .Net 4.5.2 before
   boxstarter could be used to collect the updates. Further DISM
   is not available yet for Win-2008, so the actual .iso build and
   offline servicing had to be done on a Windows-2012R2 machine.
2. The base build assumes Powershell has been injected into the image
   through the slipstream process which allows boxstarter to start
   and automatically pull in the final set of required updates.
3. Powershell scripts need to be run elevated explicitly - this is
   probably due to the UAC model for Vista/2008 SP2. But it also means
   a nasty workaround has to be employed (see windows_env.ps1) to get
   the scripts to actually work. This impacted the puppet run the most
   as it does not download/install modules unless running elevated.
4. Additional logic had to be included in several scripts to filter
   for 2008 SP2, e.g. DISM or cleanmgr is not present.
5. No attempt is made to clean WinSxS as the tools are not available
   for this release, which leads to a much larger image.
6. An extra shutdown script had to be added for the final shutdown to
   ensure the shutdown happens in a controlled manner. This seems to
   be due to the "standard" script not running in elevated mode.
7. ESX 5.5 (vSphere) doesn't play nicely with EFI for Win-2008, so have
   switched to BIOS firmware.
8. Cygwin setup needs to be called with start-process -wait to ensure it
   completes successfully, so took opportunity to refactor cygwin
   and 7zip calls to use start-process for consistency.

### DIFF
--- a/scripts/windows/clean-disk-dism.ps1
+++ b/scripts/windows/clean-disk-dism.ps1
@@ -4,8 +4,6 @@ $ErrorActionPreference = 'Stop'
 
 $SpaceAtStart = [Math]::Round( ((Get-WmiObject win32_logicaldisk | where { $_.DeviceID -eq $env:SystemDrive }).FreeSpace)/1GB, 2)
 
-$WindowsVersion = (Get-WmiObject win32_operatingsystem).version
-
 #Set all CleanMgr VolumeCache keys to StateFlags = 0 to prevent cleanup. After, set the proper keys to 2 to allow cleanup.
 $SubKeys = Get-Childitem HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\VolumeCaches
 Foreach ($Key in $SubKeys)
@@ -16,7 +14,12 @@ Foreach ($Key in $SubKeys)
 # Cleanup Windows Update area after all that
 # Clean the WinSxS area - actual action depends on OS Level - full DISM commands only available from 2012R2 and later.
 Write-Host "Cleaning up WinxSx updates"
-if ($WindowsVersion -eq '6.1.7601') {
+$WindowsVersion = (Get-WmiObject win32_operatingsystem).version
+If ($WindowsVersion -eq '6.0.6002') {
+  Write-Host "Windows 2008 - Reduced cleanup"
+  compcln /quiet
+}
+ElseIf ($WindowsVersion -eq '6.1.7601' ) {
   # Windows 2008R2/Win-7 - just set registry keys for cleanmgr utility
   reg.exe ADD "HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\VolumeCaches\Update Cleanup"       /v $CleanMgrStateFlags /t REG_DWORD /d $CleanMgrStateFlagClean /f
   reg.exe ADD "HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\VolumeCaches\Service Pack Cleanup" /v $CleanMgrStateFlags /t REG_DWORD /d $CleanMgrStateFlagClean /f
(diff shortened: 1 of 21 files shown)


scripts/windows/clean-disk-dism.ps1
```diff
@@ -30,29 +32,34 @@ ElseIf ($WindowsVersion -eq '6.2.9200') {
...
-reg.exe ADD "HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\VolumeCaches\Windows Error Reporting Archive Files"        /v $CleanMgrStateFlags /t REG_DWORD /d $CleanMgrStateFlagClean /f
-reg.exe ADD "HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\VolumeCaches\Windows Error Reporting Queue Files"          /v $CleanMgrStateFlags /t REG_DWORD /d $CleanMgrStateFlagClean /f
-reg.exe ADD "HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\VolumeCaches\Windows Error Reporting System Archive Files" /v $CleanMgrStateFlags /t REG_DWORD /d $CleanMgrStateFlagClean /f
-reg.exe ADD "HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\VolumeCaches\Windows Error Reporting System Queue Files"   /v $CleanMgrStateFlags /t REG_DWORD /d $CleanMgrStateFlagClean /f
-reg.exe ADD "HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\VolumeCaches\Windows Upgrade Log Files"                    /v $CleanMgrStateFlags /t REG_DWORD /d $CleanMgrStateFlagClean /f
+If ($WindowsVersion -eq '6.0.6002') {
+  Write-Host "Skipping CleanMgr for Windows 2008"
+}
+else {
+  # Set registry keys for all the other cleanup areas we want to address with cleanmgr - fairly comprehensive cleanup
```
Iristyle: Why not refactor this to its own helper function?

scripts/windows/windows-env.ps1
```diff
@@ -3,6 +3,26 @@ $ErrorActionPreference = 'Stop'
 
 # TODO Define variables in later tickets.
 
+# Windows Version is used for OS detection in several places, so put here.
+$WindowsVersion = (Get-WmiObject win32_operatingsystem).version
+
+If ($WindowsVersion -eq '6.0.6002') {
```
Iristyle: I think that this stdout / stderr redirection should probably be broken out into its own (well-named) function.

And that reflection, fun!

jcoconnor: Agree with the fun bit.

scripts/windows/clean-disk-dism.ps1
```diff
@@ -4,8 +4,6 @@ $ErrorActionPreference = 'Stop'
 
 $SpaceAtStart = [Math]::Round( ((Get-WmiObject win32_logicaldisk | where { $_.DeviceID -eq $env:SystemDrive }).FreeSpace)/1GB, 2)
 
-$WindowsVersion = (Get-WmiObject win32_operatingsystem).version
```
Iristyle: The global variable approach of automatically having `$WindowsVersion` defined because `A:\windows-env.ps1` has been called, is a little more difficult to maintain than either:

* Just calling it like you are here
* Passing it in as a value to functions that need it

You're not really taking a performance hit if you have to call `(Get-WmiObject win32_operatingsystem).version` a few times, and it makes the code more readable IMHO.

scripts/windows/clean-disk-dism.ps1
jcoconnor: Have refactored into hopefully more readable format. Didn't think a function was necc here.

scripts/windows/clean-disk-dism.ps1
jcoconnor: @Iristyle I'm inclined to leave this as is since:
1. The WindowsVersion is used through several of these scripts.
2. I have also deliberately placed it at the very start of the windows-env.ps1 script so as to ensure the reflection code is pulled in at the start of the script if required for win-2008.

scripts/windows/windows-env.ps1
jcoconnor: Am minded to leave this as is as the article in question advises putting it as early as possible in the script. Agree its a kludge, but unfortunately a necessary one for the snowflake that win-2008 is. (also see comment below).

jcoconnor:
This image is now up and working on ```vmpooler-dev```
I have run the puppet acceptance tests and it passes all except these two:
1. Failed to find expected binary 'C:/Program Files (x86)/Puppet Labs/Puppet/bin/puppet.bat'
2. tests/resource/file/ticket_6448_file_with_utf8_source.rb
These failures have also been seen on the other OS's and will be investigated separately.

Ping @puppetlabs/agent-platform-team @glennsarti

branan:
Does it make sense to look for "6.0.*" instead of just "6.0.6002"? Is it likely/possible that this code will ever be used without SP2? (My guess is "no", I'm just double-checking)

scripts/windows/clean-disk-dism.ps1
```diff
@@ -16,7 +14,12 @@ Foreach ($Key in $SubKeys)
 # Cleanup Windows Update area after all that
 # Clean the WinSxS area - actual action depends on OS Level - full DISM commands only available from 2012R2 and later.
 Write-Host "Cleaning up WinxSx updates"
-if ($WindowsVersion -eq '6.1.7601') {
+$WindowsVersion = (Get-WmiObject win32_operatingsystem).version
+If ($WindowsVersion -eq '6.0.6002') {
```
branan: There are seven occurrences of "6.0.6002" that are added in this diff. This is giberish to anyone not familiar with Windows versions. Could it be given a sane name (Perhaps `Server2k8sp2` in `windows-env.ps1`?) that we can use in all locations

Same goes for all other numeric Windows release versions, but this is a good place to start since we're using it here

scripts/windows/clean-disk-dism.ps1
branan: this looks better to me, I guess. Much easier to read what we're changing (even if I don't know what it all means as a non-windows person)

scripts/windows/cleanup-host.ps1
```diff
@@ -38,8 +38,14 @@ if ($env:ChocolateyToolsRoot -ne '' -and $env:ChocolateyToolsRoot -ne $null) { R
 reg.exe delete "HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment" /v "ChocolateyInstall" /f
 
 # Run Cleanmgr again.
-Write-Host "Running CleanMgr with Sagerun:$CleanMgrSageSet"
-Start-Process -Wait "cleanmgr" -ArgumentList "/sagerun:$CleanMgrSageSet"
+$WindowsVersion = (Get-WmiObject win32_operatingsystem).version
```
branan: Why doesn't this file `windows-env.ps1` to get this value?

scripts/windows/init/vmpooler-arm-host.ps1
```diff
@@ -4,6 +4,24 @@
 
 $ErrorActionPreference = 'Stop'
 
+$WindowsVersion = (Get-WmiObject win32_operatingsystem).version
```
branan: As above, it would be fantastic to have the windows version query and a bunch of symbolic names for known versions be done centrally, so that we can use human-readable names in place of numeric versions whenever possible.

scripts/windows/install-win-packages.ps1
```diff
@@ -13,51 +13,44 @@ if ("$ARCH" -eq "x86") {
   $PuppetMSIUrl = "https://downloads.puppetlabs.com/windows/puppet-agent-x64-latest.msi"
 }
 Download-File "$PuppetMSIUrl" $PackerDownloads\puppet-agent.msi
-Start-Process -Wait "msiexec" -ArgumentList "/i $PackerDownloads\puppet-agent.msi /qn /norestart PUPPET_AGENT_STARTUP_MODE=manual"
+Start-Process -Wait "msiexec" -PassThru -NoNewWindow -ArgumentList "/i $PackerDownloads\puppet-agent.msi /qn /norestart PUPPET_AGENT_STARTUP_MODE=manual"
```
branan: The `-PassThru -NoNewWindow` flags got added to many locations - does it make sense to make this something like `$WinVerSpecificMsiOptions` (name up for debate) where we can have a central comment about why they're needed in 2k8?

scripts/windows/install-win-packages.ps1
```diff
@@ -13,51 +13,44 @@ if ("$ARCH" -eq "x86") {
...
-& $7zip x C:\Packer\Downloads\autologon.zip -y $ostring
+$SysInternalsTools = @(
+  'procexp',
+  'procmon',
+  'pstools',
+  'sdelete',
+  'bginfo',
+  'autologon'
+)
+$SysInternalsTools | % {
```
branan: This is a great cleanup, though in the future changes like this should be a separate commit (not necessarily a separate PR), for the sake of making blame/history easier to understand

scripts/windows/windows-env.ps1
```diff
@@ -3,6 +3,25 @@ $ErrorActionPreference = 'Stop'
 
 # TODO Define variables in later tickets.
 
+# Windows Version is used for OS detection in several places, so put here.
+$WindowsVersion = (Get-WmiObject win32_operatingsystem).version
+If ($WindowsVersion -eq '6.0.6002') {
```
branan: More windows-version-related stuff that I wish was centralized somewhere

templates/windows-2008/README.md
```diff
@@ -0,0 +1,16 @@
+# puppetlabs-packer
+
+### About
+
+This is the Windows 2012 Packer Template
```
branan: 2008

templates/windows-2008/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+### VM settings
+
+
+## Documentation
+
+The Confluence Documentation for the process is at [Windows/Packer Imaging Process](https://confluence.puppetlabs.com/display/QE/Packer+Generation+of+Windows+Templates+for+VMPooler)
```
branan: Kind of bothers me that our documentation for a public repo is in confluence, but that is outside the scope of this PR

branan:
@jcoconnor I'm not particularly familiar with either Packer or Powershell, so take all my nit-pick-y things with a grain of salt - I'm applying general heuristics about what would make this easier for me as an outsider to read, and have no idea what's idiomatic/normal in this code

scripts/windows/install-win-packages.ps1
glennsarti: Can use parameter splatting for that.

McdonaldSeanp:
@branan had some good comments about the windows version variables, but overall I am 👍 with the actual working code.

scripts/windows/install-win-packages.ps1
jcoconnor: Yeah - I've redone the commits completely rather than as a single squashed commit.

scripts/windows/install-win-packages.ps1
jcoconnor: @branan These parameters will work for any version of windows - main reason for the PassThru is so that the Cygwin install returns its output directly to the console (and this is useful information if there are issues) - so I decided to use a consistent set for all the installs.
However, @glennsarti makes a useful point about splatting.

scripts/windows/init/vmpooler-arm-host.ps1
jcoconnor: Good idea - will implement as a MAINT commit as suggested.

scripts/windows/cleanup-host.ps1
jcoconnor: Perhaps @Iristyle would like to comment :) ?
In any case, I'm going to redo the version number detection and use the shortened numbers [ as suggested in this article](https://msdn.microsoft.com/en-gb/library/windows/desktop/ms724832%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396)

templates/windows-2008/README.md
jcoconnor: Oops

templates/windows-2008/README.md
jcoconnor: Yes - that bothers me too a bit - until recently there has been little packer documentation - the Confluence guide was more intended as internal to explain the complexities w.r.t. Windows. I think @ScottGarman has added similar Linux docs.

jcoconnor:
@branan @McdonaldSeanp @glennsarti & @Iristyle 

I've done a bit of tidying again to pickup some recent review comments:
1. PR is now 7 distinct commits to capture distinct changes that can be grouped together and allow for other refactoring such as the windows process calls.
2. There was some variance in opinion between @branan and @Iristyle as to where the $WindowsVersion should be evaluated - I have decided to put in ```windows-env.ps1```.
3. Really liked @branan's suggestion for symbols for the windows versions - hope the use of ```-like``` is acceptable here, but it does appear to improve the clarity. I have switched the ordering of the tests to avoid using ```-notlike```

I have run this through local builds on my machine and don't see any issues.
I will run it through a full build of all the OS's - this will take 1-2 days.
However, I think this is good to go.

scripts/windows/install-win-packages.ps1
```diff
@@ -12,52 +12,53 @@ if ("$ARCH" -eq "x86") {
...
 }
+
+# Define common Start-Process params appropriate for running the install setups.
+# Main one is -Wait (until setup is complete).
+# PassThru and NoNewWindow also relevant to ensure any installer console output is properly captured
+$SprocParms = @{'Wait'=$true;
+                'PassThru'=$true;
+                'NoNewWindow'=$true
+}
+
```
jcoconnor: Splatting as recommended by @glennsarti

McdonaldSeanp:
Cleanup looks very good, 👍

branan:
This looks much clearer to my eyes now. Thanks @jcoconnor :+1:

jcoconnor:
Tx team - Testing shows an issue with the ```win-2012-x86_64``` build. The ```Start-Process -Wait``` command is failing on the ```7zip``` operations. Trying a workaround and will test this on all platforms first.

**PR set to Don't Merge.**

jcoconnor:
I've tested the build and deploy to ```vmpooler-dev``` on all 12 platforms.

The only error encountered was with with ``win-10-ent-i386``` - the image doesn't sysprep. After initial investigation, this appears to be a ```win-10-i386``` specific issue - the x86_64 platform is ok that may be due to the most recent Windows update, so a separate ticket [IMAGES-503](https://tickets.puppetlabs.com/browse/IMAGES-503) has been created for this.

Puppet acceptance tests were run on the following platforms:
1. ```win-2008-x86_64``` (the new platform)
2. ```win-2012r2-x86_64``` (existing main build platform)
Testing passed with the exception of 2 known failures that are not related to the new platforms.

So this PR is now ready for merge.

Ping @glennsarti @puppetlabs/agent-platform-team

PR #187, opened 2017-03-29
corrects faulty ignore of packer manifests files

witjoh:
Think this was introduced while adding tests.  
manifests/modules/packer/* was also ignores, which seems not correct.


shrug:
Thanks!

PR #192, opened 2017-04-21
Updates for SLES-12 sp1 and sp2 builds

shrug:
This adds a step to make sure sles isn't adding locks to packages and updates the template names to be consistent between sp1 and sp2

### DIFF
--- a/manifests/modules/packer/templates/vsphere/sles.rb.erb
+++ b/manifests/modules/packer/templates/vsphere/sles.rb.erb
@@ -24,6 +24,7 @@ Kernel.system('systemctl restart wicked.service')
 puts '- Cleaning up...'
 
 Kernel.system('rm /etc/vsphere-bootstrap.rb')
+Kernel.system('rm /etc/zypp/repos.d/SLES*.repo')
 Kernel.system('echo "exit 0" > /etc/rc.d/after.local')
 
 puts "\n"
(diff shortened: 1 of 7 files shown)


shrug:
Running test builds now...

manifests/modules/packer/templates/vsphere/sles.rb.erb
```diff
@@ -24,6 +24,7 @@ Kernel.system('systemctl restart wicked.service')
 puts '- Cleaning up...'
 
 Kernel.system('rm /etc/vsphere-bootstrap.rb')
+Kernel.system('rm /etc/zypp/repos.d/SLES*.repo')
```
shrug: This is a repo definition for the install media, so when you try to install something it tells you to insert the disc... I imagine there's a way to get the autoyast install to not create this, or to leave it disabled but I haven't worked out how, and I'm wasting time messing with it. So here we are.

templates/sles-12.1/files/sles-12-sp1-x86_64-autoinst.xml
```diff
@@ -127,7 +127,7 @@
         <partition>
           <create config:type="boolean">true</create>
           <crypt_fs config:type="boolean">false</crypt_fs>
-          <filesystem config:type="symbol">btrfs</filesystem>
+          <filesystem config:type="symbol">ext4</filesystem>
```
underscorgan: -_- was this the issue?

shrug: yuuuup

PR #195, opened 2017-05-08
(IMAGES-539) Disable Google Update Services

jcoconnor:
The Udpate services cause a pending reboot request which interferes with
module testing, so disable this.


jcoconnor:
Closing as these changes are merged into #196

PR #197, opened 2017-05-19
(MAINT) Try using choco .net4.5.2

jcoconnor:
This seems to be a preferable choco package for the .net install.
Apply it across builds except Win-2008.


jcoconnor:
Closing as #198 incorporates this maint fix.

PR #200, opened 2017-06-07
(PE-20580) Remove yum-autoupdate package from SL6

shrug:


stahnma:
Should we do this on all other el-based templates?

shrug:
This package appears to be a scientific thing, el/centos, etc, use a different mechanism (that e don't enable)

PR #203, opened 2017-07-31
(IMAGES-569) Update debian 9 image to released version, add hack

McdonaldSeanp:
This commit adds the released stable debian 9 images, and adds a hack to
install jessie libreadline6 packages so the debian jessie agent works on
9 (and we can subsequently use it to provision the VM).

When we release the official version of the debian 9 agent, this hack should
be removed

### DIFF
--- a/templates/debian-9.0/x86_64.vmware.base.json
+++ b/templates/debian-9.0/x86_64.vmware.base.json
@@ -6,8 +6,8 @@
       "template_os": "debian7-64",
       "template_config": "base",
 
-      "iso_url": "http://osmirror.delivery.puppetlabs.net/iso/debian-stretch-DI-rc3-amd64-xfce-CD-1.iso",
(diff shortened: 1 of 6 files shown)


templates/debian-9.0/x86_64.vmware.base.json
```diff
@@ -3,20 +3,20 @@
   "variables":
     {
       "template_name": "debian-9.0-x86_64",
-      "template_os": "debian7-64",
+      "template_os": "other",
```
stahnma: Why is this not debian9-64 or something of the like?

McdonaldSeanp: Oh, yeah, I should change it back

ScottGarman:
This looks good, I assume you'll be adding the i386 arch as well?

McdonaldSeanp:
Yeah, working on that now

PR #205, opened 2017-08-01
(maint) Add virtual devs and update if statements to use Major release

McdonaldSeanp:


ScottGarman:
:+1:

manifests/modules/packer/manifests/vmtools.pp
```diff
@@ -2,8 +2,7 @@
 
   case $::osfamily {
     debian: {
-      # Debian 9.0 alpha release returns 'stretch/sid' for operatingsystemrelease:
-      if ($::operatingsystemrelease in ['stretch/sid']) or ($::operatingsystemmajrelease in ['7', '8', '16']) {
+      if ($::operatingsystemmajrelease in ['7', '8', '9', '16']) {
```
branan: I thought we fixed facter to return full year.month versions for ubuntus? Is the version of puppet we're using for this old enough to not have them? Or is that '16' just not that useful anymore?

McdonaldSeanp: I think the 16 just isn't useful anymore.

This is something we need to keep in mind when we update the Ubuntu images next.

branan: Cool. Should we rip it out as part of this, or as a separate cleanup PR?

McdonaldSeanp: A separate cleanup PR I think

branan: Works for me :+1:

PR #208, opened 2017-08-09
(IMAGES-596) Fix Sysprep/Winapp issue

jcoconnor:
The latest cumulative update to Windows-10 seems to add or re-enable a number of the Metro apps which are causing a sysprep failure. Revise the App cleanup to remove all Metro Apps and to repeat the exercise after the Windows Update cycle is complete.

Also fix the following two issues which were impacting this Patch Tuesday general refresh of images:
1. Add sleep to Powershell to overcome a Packer/Powershell intermittent issue
2. Persist drivers in the Post-Clone Sysprep - should reduce pooler instance startup time.

### DIFF
--- a/scripts/windows/windows-env.ps1
+++ b/scripts/windows/windows-env.ps1
@@ -303,37 +303,44 @@ Function Install-PackerWindowsUpdates
 
 Function Remove-Win10Packages
 {
-  if (-not (Test-Path "A:\RemoveWin10Packages.installed"))
-  {
-    Write-Host "Remove All Win-10 App/Packages to prevent Sysprep Issues"
+  Write-Host "Remove All Win-10 App/Packages to prevent Sysprep Issues"
 
-    Import-Module Appx
-    Import-Module Dism
+  Import-Module Appx
+  Import-Module Dism
 
-    Get-AppxPackage | % {if (!($_.IsFramework -or $_.PublisherId -eq "cw5n1h2txyewy")) {$_}} | Remove-AppxPackage -ErrorAction SilentlyContinue
-    Get-AppXProvisionedPackage -Online | Remove-AppxProvisionedPackage -Online -ErrorAction SilentlyContinue
+  Write-Host "Removing AppxPackages"
+  Get-AppxPackage -AllUsers | Remove-AppxPackage -ErrorAction SilentlyContinue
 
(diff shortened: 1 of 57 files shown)


jcoconnor:
This is ready for merge - Both Win-10 images build and run thru acceptance suite without issue (see build log links below).

In addition, the Sysprep with ```PersistAllDeviceInstalls``` is building and running through ```sysprep``` correctly in all acceptance tests. The complete suite of passing Windows Builds can be viewed at: https://jenkins-imaging.delivery.puppetlabs.net/view/Windows/

https://jenkins-imaging.delivery.puppetlabs.net/view/Windows/job/imaging_windows-10_x86_64_vmware_vsphere.cygwin_acceptance_packer/19/console

https://jenkins-imaging.delivery.puppetlabs.net/view/Windows/job/imaging_windows-10_i386_vmware_vsphere.cygwin_acceptance_packer/15/console

/cc: @McdonaldSeanp & @Iristyle

Iristyle:
The `Start-Sleep` part seems really unfortunate.  Everything looks fine, but wish we had a solution other than "sleep and hope it works"

jcoconnor:
Agree on the ```Start-Sleep``` - will continue to follow the issue on the Packer Issues list - I think a bug really needs to be filed rather than a question.

jcoconnor:
Will see if I can do a better workaround.

jcoconnor:
The packer discussion has come up with another workaround that I'm going to put through instead.
See: https://github.com/hashicorp/packer/issues/5064#issuecomment-322973639

```
{
"type": "windows-restart",
"restart_command": "shutdown -r -f -t 10 && sc config winrm start= delayed-auto && net stop winrm",
"restart_timeout": "10m"
}
```

jcoconnor:
@Iristyle There has been further discussion on https://github.com/hashicorp/packer/issues/5064 and a recommendation to try specifying a ```remote_path``` for the target powershell script.

I was thinking of doing this anyway for traceability.
Have tested the workaround and it appears to resolve the problem for now.

Would like to get this merged so that I can get down to some refactoring on the overall scripts including:
1. Use common json files across most of the windows builds with platform specific variable files to substantially reduce the copy/pasta.
2. Replacing boxstarter across all the builds.

/cc @McdonaldSeanp

Iristyle:
This solution looks better.

PR #211, opened 2017-08-11
(IMAGES-595) Remove debian 9 libreadline hack

McdonaldSeanp:
This commit removes the hack we used to get debian 9 to work by using the
newly released agent 1.10.6 for debian 9


ScottGarman:
:+1:

PR #212, opened 2017-08-14
(IMAGES-592) Update configuration, repos for fedora 26

McdonaldSeanp:
This commit does 2 things:

1) Updates several parts of the packer configuration that were outdated and
causing problems in the imaging pipeline

2) Updates the puppet manifests for fedora packages to correctly use fedora26
instead of fedora-26 in the URL for mirrored repos

### DIFF
--- a/templates/fedora-26/x86_64.vmware.base.json
+++ b/templates/fedora-26/x86_64.vmware.base.json
@@ -4,6 +4,8 @@
     {
       "template_name": "fedora-26-x86_64",
       "template_os": "fedora-64",
+      "template_config": "base",
+
 
       "iso_url": "http://osmirror.delivery.puppetlabs.net/iso/Fedora-Server-dvd-x86_64-26-1.5.iso",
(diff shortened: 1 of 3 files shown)


templates/fedora-26/x86_64.vmware.base.json
```diff
@@ -49,7 +52,7 @@
       "tools_upload_flavor": "linux",
       "vmx_data": {
         "cpuid.coresPerSocket": "1",
-	"ethernet0.pciSlotNumber": "32",
+	      "ethernet0.pciSlotNumber": "32",
```
smcclellan: Spacing here looks off?

McdonaldSeanp: Huh, it looked off on my machine, that's why I made that change.

Suuuuuper weird

McdonaldSeanp: AH, there was a tab there before.

PR #213, opened 2017-08-21
(IMAGES-600) Slipstream Helper Scripts

jcoconnor:
These scripts were used on a Win-2012r2 machine to generate the final
slipstream image using the CAB files grabbed from the win-2008/2008r2/7
builds as the DISM commands on these platforms didn't reliably generate
the slipstream disk.


McdonaldSeanp:
👍

PR #214, opened 2017-08-21
(IMAGES-594) Add 'beakerhost' to fedora 26 vsphere image

McdonaldSeanp:
the 'beakerhost' variable defines what string to give beaker-hostgenerator to
produce the configuration for this platform. This commit updates the fedora26
config to include that var


ScottGarman:
:+1: will merge when travis is done.

PR #217, opened 2017-10-23
(IMAGES-639) Remove Store/App packages for win-8.1

jcoconnor:
Applies the same method that was used for Windows-10 to resolve Sysprep
issues with the store/apps packages by completely removing them as they
are not needed for vmpooler etc.

Rename method for Win10 to more generic name.

### DIFF
--- a/scripts/windows/windows-env.ps1
+++ b/scripts/windows/windows-env.ps1
@@ -295,9 +295,9 @@ Function Install-DotNetLatest
 }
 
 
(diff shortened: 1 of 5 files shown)


jcoconnor:
I have successfully built and tested this using the Jenkins-Imaging pipelines.
The image boots and syspreps correctly when cloned as a new VM and runs the Packer/Imaging acceptance tests.

So Ready for Review/Merge.

jcoconnor:
This is merged into #218

PR #221, opened 2017-10-27
(IMAGES-614) Fix BHG Defs for Core

jcoconnor:
Need a _ instead of - to seperate core for the Beaker Host Generator
Defs for Win-2012r2-core and win-2016-core acceptance tests


McdonaldSeanp:
👍

PR #223, opened 2017-11-07
(IMAGES-656) Create a RHEL7 FIPS mode packer image

jsane:
Created new rhel7 template with required scripts to configure it to
  - Enable repo's to be able install packages.
  - Boot in FIPS mode and
  - Install openssl package version aligned with that of openssl-devel

This is installing the currently available version 1.9.3 of puppet agent.
Repo enabling is being done using a script which in future could be
reworked to be done via puppet manifest (vsphere/repos.pp)
Note: Step for loading the image in vsphere is still remaining.

### DIFF
--- a/manifests/modules/packer/manifests/networking/params.pp
+++ b/manifests/modules/packer/manifests/networking/params.pp
@@ -7,38 +7,44 @@ class packer::networking::params {
     }
 
     redhat: {
-      case $::operatingsystemrelease {
-        '7.0.1406', '7.1.1503', '7.2.1511', '7.2': {
-          case $::provisioner {
-            'virtualbox','libvirt': { $interface_script = '/etc/sysconfig/network-scripts/ifcfg-enp0s3' }
-            'vmware':     { $interface_script = '/etc/sysconfig/network-scripts/ifcfg-ens33' }
-          }
 
-          $udev_rule     = '/etc/udev/rules.d/70-persistent-net.rules'
-          $udev_rule_gen = '/lib/udev/rules.d/75-persistent-net-generator.rules'
-        }
+       case $::operatingsystemrelease {
+         '7.0.1406', '7.1.1503', '7.2.1511', '7.2': {
+           case $::provisioner {
+             'virtualbox','libvirt': { $interface_script = '/etc/sysconfig/network-scripts/ifcfg-enp0s3' }
+             'vmware':     { $interface_script = '/etc/sysconfig/network-scripts/ifcfg-ens33' }
+           }
+ 
+           $udev_rule     = '/etc/udev/rules.d/70-persistent-net.rules'
+           $udev_rule_gen = '/lib/udev/rules.d/75-persistent-net-generator.rules'
+         }
 
-        '5.11': {
-          $interface_script = '/etc/sysconfig/network-scripts/ifcfg-eth0'
-          $udev_rule        = '/etc/udev/rules.d/70-persistent-net.rules'
-        }
-
-        '23', '24', '25', '26': {
-          case $::provisioner {
-            'virtualbox': { $interface_script = '/etc/sysconfig/network-scripts/ifcfg-enp0s3' }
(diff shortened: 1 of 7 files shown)


manifests/modules/packer/manifests/networking/params.pp
```diff
@@ -34,7 +41,7 @@
         }
 
         default: {
-          $interface_script = '/etc/sysconfig/network-scripts/ifcfg-eth0'
+          $interface_script = '/etc/sysconfig/network-scripts/ifcfg-ens33'
```
ScottGarman: Curious why you're changing the default here? Is it because most new OS releases going forward are going to use ens33 instead of eth0 as the default interface name?

jsane: Agree it should not be there (in default). But the "case $::operatingsystemrelease" does not seem to resolve to '7' where I could have added that. I then attempted adding an "if $::operatingsystemmajrelease in ['7']" at various places within the case but got flagged as syntax error (may be lack of my ruby fluency). Am going to try one more thing and see how it goes.

jsane: This should be resolved with my last push. I notice that this rhel7 image has its network interface as ens33 while the ones in use in vmpooler have ens192. Am guessing that is ok.

ScottGarman:
:+1:  I'm willing to go with this for now and make additional tweaks as necessary.

manifests/modules/packer/manifests/networking/params.pp
```diff
@@ -7,36 +7,45 @@
...
-          case $::provisioner {
-            'virtualbox': { $interface_script = '/etc/sysconfig/network-scripts/ifcfg-enp0s3' }
-            'libvirt': { $interface_script = '/etc/sysconfig/network-scripts/ifcfg-ens4' }
-            'vmware':     { $interface_script = '/etc/sysconfig/network-scripts/ifcfg-ens32' }
-          }
-          $udev_rule     = undef
-          $udev_rule_gen = undef
-        }
-
         default: {
```
shrug: So I think this default will never get hit, since "$::operatingsystemmajrelease" will always resolve to '7' for these

scripts/enable-rhel-repos.sh
```diff
@@ -0,0 +1,35 @@
+
+# Script to provision Red Hat repos
```
shrug: Should this be handled in https://github.com/puppetlabs/puppetlabs-packer/blob/master/manifests/modules/packer/manifests/vsphere/repos.pp instead of a separate script?

manifests/modules/packer/manifests/networking/params.pp
jsane: ok... Since we are using vmware am now thinking if I even need to have that check on operatingsystemmajrelease since am trying to configure the same network i/f as that. Will check. thx

scripts/enable-rhel-repos.sh
jsane: I think I had attempted that earlier but was running into some issues.. will check again but may need some help with it.

jsane: ok I remember now. So per the current implementation in repos.pp it should have had configured at least two repo's for os and updates for redhat but none gets configured. 
So I retraced/tried again and even w/o introducing anything that I added for enabling the repos etc. the base line template + all existing manifests it cannot install any packages it attempts from bootstrap-aio.sh script (because there are no repos). 
While I agree that fixing the repos.pp would the proper way to do this but this is getting beyond my expertise - I don't understand why the existing repos.pp won't configure the two repos.

manifests/modules/packer/manifests/networking/params.pp
jsane: Pl let me know the latest version looks ok.

ScottGarman:
Closing this in favor of https://github.com/puppetlabs/puppetlabs-packer/pull/245 which I refactored to use the new imaging template structure.

PR #229, opened 2017-12-08
(IMAGES-677) Cleanup templates for Debian platforms

ekinanp:


ScottGarman:
This looks pretty good. Maybe bump the template version numbers, since we're changing the default ethernet device?

ekinanp:
More of a reminder to myself, but it seems that VMWare Fusion 8.5.x does not support Debian 9, so the guest OS for those templates are still Debian 8 (build failed when I set it to Debian 9). Need to check if this is still the case once I upgrade to VMWare Fusion 10.

ekinanp:
Ok, I have updated this PR with the new directory structure. The main thing to note is that in the common templates, variables with the value "null" are what the template requires in the corresponding variables file in order to run successfully. All other variables there have defaults and can be overridden in the variables file.

ekinanp:
I have removed the CHANGELOGs for the Debian platforms because the Windows templates do not have them. I can add them back if needed.

The Windows templates do have READMEs which I can add to the Debian templates as well if desired.

@jcoconnor @ScottGarman @heathseals What do you all think?

jcoconnor:
@ekinanp This looks good - happy to merge it as-is, as I reckon we will have a few cycles needed to get all the linux build "back on air".

Please add the READMES - I am have a ticket open to update all the windows READMES.

I have some ideas on versioning that I'm going to float on a separate e-mail that I think will help the auto-deploy process.

heathseals:
@ekinanp the changelogs are referenced within the description of the boxes we've shipped to vagrantcloud in the past, [debian-7.8-64-nocm](https://app.vagrantup.com/puppetlabs/boxes/debian-7.8-64-nocm) for example.  not sure if there has been any discussion around reviving the practice of shipping artifacts to vagrantcloud, but just wanted to mention that the changelogs are linked from there.

ekinanp:
@heathseals Ah ok, sounds good. Thank you. Then I think it might be more useful to include the CHANGELOGS if we find that we need to ship to vagrantcloud again (b/c we can always get them back via. version control). @ScottGarman What do you think?

ekinanp:
@jcoconnor I added a single README for Debian itself. If you like, I can add READMEs to each of the variants (7.8, 8.2, 9.0) but I feel like that might be redundant.

templates/debian/README.md
```diff
@@ -0,0 +1,16 @@
...
+### About
+
+This contains templates for Debian platforms. Currently, only Debian 7.8, 8.2 and 9.0 are supported for both the 64-bit and 32-bit architectures.
+
+### VM settings
+
+
+## Documentation
+
+The Confluence Documentation for the process is at [Linux/Packer Imaging Process](https://confluence.puppetlabs.com/display/SRE/Linux+Image+Packer+Generation)
```
jcoconnor: @ekinanp I like the idea of a common README
Was thinking of something similar with the Windows platform, with a common detailed README in the win-common directory and smaller README's with a link to the common.

I have a ticket open (IMAGES-637) to document the windows build in more detail including the actual packer commands needed to build. So might be no harm if you put up a similar ticket to make sure same is done in Debian.

On other think we need to do is remove the Confluence link - its not visible outside puppet.

ScottGarman:
I'm with John on still including Readme files - especially to document why we have those null values in the templates, and to provide an example on running a packer build command that uses the var files. With the vagrant templates going away, we can remove the Changelogs.

ekinanp:
Ok, I have also verified that Debian 9 is supported on VMWare Fusion 10 so this PR should be ready for merge.

PR #230, opened 2017-12-11
(RE-9766) Update puppetlabs-packer to use artifactory

jennaesq:
This change updates references from using http://buildsources.delivery.puppetlabs.net/ to https://artifactory.delivery.puppetlabs.net/artifactory/generic/buildsources/.

### DIFF
--- a/scripts/windows/generate-slipstream.ps1
+++ b/scripts/windows/generate-slipstream.ps1
@@ -20,7 +20,7 @@ $WinImageFile = "$WinDistPath\sources\install.wim"
 
 # Install ADK
 Write-Output "Install Win ADK"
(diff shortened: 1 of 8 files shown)


jennaesq:
sed is your friend;  once you realize BSD sed is installed on your mac, rather than GNU sed. (Mac growing pains)

mwaggett:
Is there a higher level place this url could live, so future updaters would only have to update in one place? @shrug ?

shrug:
Not the way the packer stuff is currently built, each of those files is an independent script, they don't template or source parent scripts. Everything could be refactored, but that's a bigger job. :)

shrug:
ping @jcoconnor for review

jcoconnor:
Going to merge this and add the equivalent to the ISO pulls in the windows variable files.

PR #233, opened 2017-12-12
(maint) Update Travis to not fail with the new naming convention

ekinanp:
With the new directory structure of the templates established,
variable files will be named "vars.json". This would fail Travis
because it expects "vars.json" files to have a prefix, originally
meant for the architecture. This commit fixes Travis to avoid
failure for that case.

### DIFF
--- a/tests/test-syntax.sh
+++ b/tests/test-syntax.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
 
 counter=0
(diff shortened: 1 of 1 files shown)


ekinanp:
Note that the CI failures are expected -- the windows templates have not yet been updated to the new directory structure so var files still have the architecture prefixed to them.

PR #234, opened 2017-12-13
(maint) Add back old filter so that CI passes

ekinanp:
This should be reverted once everything conforms to the new
directory structure.

### DIFF
--- a/tests/test-syntax.sh
+++ b/tests/test-syntax.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 counter=0
-for file in $(find . -name "*.json" | egrep -v '^vars\.|MAINTAINERS|metadata')
+for file in $(find . -name "*.json" | egrep -v '^vars\.|.(vars|variables).|MAINTAINERS|metadata')
 do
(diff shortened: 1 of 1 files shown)


tests/test-syntax.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 counter=0
-for file in $(find . -name "*.json" | egrep -v '^vars\.|MAINTAINERS|metadata')
+for file in $(find . -name "*.json" | egrep -v '^vars\.|.vars.|MAINTAINERS|metadata')
```
jcoconnor: @ekinanp I think the regex should be:

```'^vars\.|.variables.|MAINTAINERS|metadata'```

ekinanp: @jcoconnor Updated the PR to include ".variables.". Forgot about the Centos templates resulting from the community PR. The ".vars." part should probably still be there b/c there are Windows var files that are named as <arch>.vars.json

jcoconnor:
The good indicator is that tests pass now.
👍 
Going to merge.

PR #235, opened 2017-12-13
(IMAGES-680) Win-10 Needs private network adapters

jcoconnor:
Windows-10/2016 insists on creating public network adaptors after a sysprep which means WinRM fails to come up properly (Earlier versions of windows allow you to set the adaptors private in the unattend.xml).

So detect if this Windows-10/2016 and set the adapters private.


jcoconnor:
Win-10 builds ok on Vagrant again.

This is ready for review/merge.

PR #236, opened 2017-12-15
(MAINT) Don't start sshd during post-clone

jcoconnor:
Small maint change to improve the acceptance reliability.
Starting sshd before we reboot to rename the host causes failures in the
vcloud/beaker hypervisor - suspect this is because it uses IP addresses
rather than the given hostname to detect when the host is alive, thus
breaking the test initialisation sequence. The vmpooler initialisation
is more tolerant of this as it the ssh test uses the given hostname
which isn't active until after the reboot.

Suspect we can further rationalise and tidy the overal ssh startup
sequence (i.e. use Automatic start), but lets just do this for now.


jcoconnor:
I've tested this with the acceptance test run for {{win-2008r2-x86_64}} - this appears to fix the startup sequence and allow the test to proceed.

PR #238, opened 2017-12-19
(maint) Move Linux platforms to fit new directory structure

ekinanp:


templates/linux/debian/common/vmware.base.json
```diff
@@ -0,0 +1,109 @@
+{
```
ekinanp: The only reason this was created was b/c Packer does not interpolate {.HTTPIP} and {.HTTPPort} when the boot command is declared as a user variable. I was also not able to load Debian's preseed file from the floppy disk as it could not find the preseed file -- this seems to be a known bug per https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=783982 and I do knot know of a way to fix it. Thus, if someone figures out how to preseed Debian from floppy, then this template is not needed.

Thus, any updates to the linux/common/vmware.base.json template should also be reflected here.

ScottGarman:
:+1:

PR #243, opened 2017-12-21
(IMAGES-679) Refactor Oracle Linux 5/6 image templates

ScottGarman:
This removes unmaintained templates for virtualbox and vagrant, and
applies the new var file and directory format restructuring that is
being applied to the rest of our supported Linux platforms.


ScottGarman:
@ekinanp good catch on both counts. It looks like the current pooler oracle 6 VMs don't include patch, but this is definitely something we want on all of our vms, and the other oracle VMs do have net-tools.

PR #244, opened 2017-12-21
(IMAGES-676) Cleanup templates for Ubuntu platforms

ekinanp:


ScottGarman:
My only nitpick is that you're still using puppet-agent 1.8.3, which could be updated to use 1.10.9 to be more current.

ekinanp:
@ScottGarman Yeah I noticed that you're using 1.10.9 which is what I should do. I have been using 1.8.3 for all the clean-up stuff so far I guess b/c I was worried 1.10.x wouldn't work for some reason?
 I can clean it up later once everything is finished as a separate PR.

ScottGarman:
Sure, making a pass afterward to upgrade the agent version is fine. Everything should work with the LTS agent version.

:+1:

PR #247, opened 2017-12-22
(maint) Convert the provisioning scripts into an overridable var

ScottGarman:
We'd like var files to be able to override and specify multiple
provisioning scripts.


ScottGarman:
Thanks @ekinanp, I took your advice and will merge this with that improvement.

PR #248, opened 2017-12-27
(IMAGES-684) Fix File/Directory clean

jcoconnor:
The "icacls" command was failing after it was localised - fix this and
ensure all permission changes/deletes are logged verbosely so that we
can verify the cleanup is happening (may choose to switch verbose mode
off at a later stage once this is working stably).


jcoconnor:
Have just tested this on the build: ```imaging_win-2012r2_x86_64_vmware_vsphere.cygwin_packer``` and it is correctly deleting the ```C:\ProgramData\PuppetLabs``` directory.

This likely addresses [IMAGES-671](https://tickets.puppetlabs.com/browse/IMAGES-671) as well.

PR #249, opened 2017-12-27
(IMAGES-689) Cleanup Solaris templates so that they build

ekinanp:
There are no vsphere templates for the Solaris platforms; instead,
vmware.vagrant was used. This work will be done in IMAGES-695


ekinanp:
It would have been nice to have a common template for Sol 10u11 and 11.2, but unfortunately there is no floppy support in Sol 11.2 (per http://partnerweb.vmware.com/GOSIG/Solaris_11.html), so the boot command must receive the installation preseed files from the http server. This is something that Packer cannot interpolate when the boot command is passed in as a user variable (similar to the case with the Debian platforms).

ScottGarman:
:+1:

templates/linux/solaris/10u11/common/vmware.base.json
```diff
@@ -0,0 +1,75 @@
+{
```
caseywilliams: Would there be any harm in putting this in a `10` directory instead of `10u11` (and make some sort of similar change for 11)? Just to make the versions match up between what's here and what we see when we look at the list of vmpooler templates as vmpooler users? That would make more sense to me, but it's just a suggestion - I don't know that much about the intricacies of solaris releases.

ekinanp: I don't think it would be good for Sol 11, b/c e.g. we have Centos 5.11, 7, 7.2. Not sure about 10u11 though -- is 10u11 a variant of Solaris? @ScottGarman @heathseals What do you think?

caseywilliams:
Moving these templates to a dedicated `solaris` directory was mentioned as an idea [here](https://github.com/puppetlabs/puppetlabs-packer/pull/254#discussion_r159981568) this has been blocked until the macos PR is merged, but I don't see a reason to block it since we're just talking about reorganizing the solaris directory structure.

ekinanp:
@caseywilliams Reason is mostly the docs -- you have the structure written out in https://github.com/puppetlabs/puppetlabs-packer/pull/254 so I figured there was no need to repeat it here until that one was merged in. However if there is an urgency to get this through, then I can go ahead and do it.

ScottGarman:
We can merge this, but need to coordinate with @jcoconnor about timing to make sure it doesn't interfere with his Windows pipeline runs that are ongoing.

PR #253, opened 2018-01-04
(IMAGES-707) Set SSHD as Automatic in post-clone

jcoconnor:
The SSHD startup was still happening in a seperate script to the
post-clone operation. This seems to cause a failure for sshd to come up
in the image accpeptance test env.

Simplify this by setting sshd as Automatic as one of the last commands
before the computer is renamed and take the startup out of the
startup-script.
The mkpasswd and mkgroup commands are also copied to the post-clone
operation to ensure these are in place before the first sshd
initialisation.

### DIFF
--- a/scripts/windows/init/vmpooler-post-clone-configuration.ps1
+++ b/scripts/windows/init/vmpooler-post-clone-configuration.ps1
@@ -11,6 +11,9 @@ function ExitScript([int]$ExitCode){
 	exit $ExitCode
 }
 
+# Used Frequently throughout
+$CygwinDir = "$ENV:CYGWINDIR"
+
 # Windows version checking logic is copied here as its not present by Default
 # on the installed system (might be an idea to change this in the future)
 Set-Variable -Option Constant -Name WindowsServer2008   -Value "6.0.*"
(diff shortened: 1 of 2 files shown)


jcoconnor:
I have run this through build and acceptance tests for:
- ```win-2008r2-x86_64```
- ```win-2012r2-x86_64```
- ```win-2012r2-fr-x86_64```
The ```sshd``` startup is a lot smoother for acceptance tests, as the service comes up immediately after the host name change/reboot.
There are a few errors on the French Run, but these are localisation issues (that need to be filed as tickets).

So this PR is ready to go.
/cc: @michaeltlombardi

scripts/windows/init/vmpooler-post-clone-configuration.ps1
```diff
@@ -85,16 +87,7 @@ Write-Output "Register the Cygwin LSA authentication package "
...
-Write-Output "Set SSHD Process with Manual Startup"
-Set-Service "sshd" -StartupType Manual
-
-Write-Output "Re-enable NETBios and WinRM Services"
-Set-Service "lmhosts" -StartupType Automatic
-Set-Service "netbt" -StartupType Automatic
-Set-Service "WinRM" -StartupType Automatic
-
-# Set Startup script (starts sshd)
+# Set Startup script (does very little except get the
```
michaeltlombardi: Is this comment meant to finish?

michaeltlombardi: Other than this minor nit, @jcoconnor, looks good!

jcoconnor: Umm - yes - fixed

PR #254, opened 2018-01-05
(IMAGES-624) Organize and update macOS templates, add 10.13 template

caseywilliams:
Reorganizes existing macOS templates and introduces new template for 10.13 based on the 10.12 image.

I've moved these templates from the `linux` subdirectory into a new one, `macos`, and supplied a README there describing how to build the images and the rationale behind why they're done this way.

Building the 10.13 image will require uploading a 10.13 installer image to our osmirror (I've been using a local file so far) - the proposed URL is in the template, but the file doesn't actually exist there yet.


templates/macos/10.12/x86_64/vmware.base.json
```diff
@@ -0,0 +1,81 @@
+{
```
ekinanp: It would be good to list all of the variables that this template requires (template_os, iso_checksum, etc.) like we do in e.g. https://github.com/puppetlabs/puppetlabs-packer/blob/master/templates/linux/debian/common/vmware.base.json, even if some of them (e.g. template_os) might be shared with other templates.

Hmm, the current build script does not allow for you to have a template and a vars file in the same directory, but I see that you did that b/c the template_name is used in more than one place. It is not hard to modify it to let you have both a var file and a template in the same directory, however. I'll can go ahead and do that, though some READMEs might need to be updated. @ScottGarman what do you think? The intuitive explanation is that you have a custom template, the var file is useful b/c some variables in that custom template (e.g. template_name) are used in other common templates.

templates/macos/10.13/x86_64/vmware.base.json
```diff
@@ -0,0 +1,76 @@
+{
+  "description"                                    : "Builds a macOS 10.13 High Sierra base template VM for use in VMWare",
+  "_comment"                                       : "This is a customization of the macOS 10.12 base image, so you must build that image first. In this template we just copy in the 10.13 installer and run the upgrade process. This approach is taken becuase of limitations in the installer image customization process from 10.12.4 onward. Null variables should be provided by vars.json files, or at the command line - see the macOS README for examples.",
+
+  "variables"                                      : {
+    "output_directory"                             : "{{env `PACKER_VM_OUT_DIR`}}",
+    "source_directory"                             : "{{env `PACKER_VM_SRC_DIR`}}",
+    "ssh_password"                                 : "{{env `QA_ROOT_PASSWORD`}}",
+    "template_name"                                : null,
```
ekinanp: It'd be nice to put null variables at the top, just to keep things consistent w/ how it is in other templates.

templates/macos/10.13/x86_64/vmware.base.json
```diff
@@ -0,0 +1,76 @@
...
+      "source"                                     : "{{user `installer_url`}}",
+      "destination"                                : "/private/tmp/high-sierra-installer.tar.gz"
+    },
+    {
+      "type"                                       : "shell",
+      "execute_command"                            : "chmod +x {{ .Path }}; sudo {{ .Vars }} {{ .Path }}",
+      "expect_disconnect"                          : true,
+      "inline"                                     : [
+        "tar -xvzf /private/tmp/high-sierra-installer.tar.gz -k -C /",
+        "\"/Applications/Install macOS High Sierra.app/Contents/Resources/startosinstall\" --agreetolicense --nointeraction --converttoapfs NO"
```
ekinanp: Not familiar with this stuff, but would the existing 10.12 tools still be 10.13? Tools like xcode-cli, for example?

templates/macos/common/vars.json
```diff
@@ -0,0 +1,15 @@
+{
+  "_comment"                : "Variable values common to all macOS versions (see <version>/<arch>/vars.json for version-specific values)",
+  "common_files"            : "./common/files",
+  "disk_size"               : "40960",
+  "headless"                : "true",
+  "install_xcode_cli_tools" : "false",
+  "iso_url"                 : "http://osmirror.delivery.puppetlabs.net/iso/OSX_InstallESD_10.12_16A312a.dmg",
```
ekinanp: I think most of this should be in the 10.12 template, at least the iso specific information and xcode_cli stuff, since it is used there only. The version value should definitely be in the corresponding arch. vars.json file.

templates/macos/common/vmware.vsphere.nocm.json
```diff
@@ -0,0 +1,91 @@
...
+  "description"                             : "Builds a vSphere template using an existing base macOS template build",
+  "_comment"                                : "This is a common vmware vsphere template for macOS platforms; It customizes a macOS 10.12 or 10.13 base image for use with vmpooler. Null variables should be provided by vars.json files within the version-specific subdirectories, or at the command line",
+  "variables": {
+    "template_name"                         : null,
+    "beakerhost"                            : null,
+    "version"                               : null,
+    "output_directory"                      : "{{env `PACKER_VM_OUT_DIR`}}",
+    "source_directory"                      : "{{env `PACKER_VM_SRC_DIR`}}",
+
+    "ssh_username"                          : "{{user `ssh_username`}}",
```
ekinanp: This won't work, unfortunately Packer does not let you interpolate user variables in other user variables. Though I think you meant to take this out.

templates/macos/common/vmware.vsphere.nocm.json
```diff
@@ -0,0 +1,91 @@
...
+        "SSH_USERNAME={{user `ssh_username`}}"
+      ],
+      "scripts"                             : [
+        "{{user `common_files`}}/pooler-setup.sh"
+      ]
+    }
+  ],
+
+  "post-processors"                         : [
+    {
```
ekinanp: You can comment/delete this part out if you want to test that your vsphere template builds.

templates/macos/10.13/x86_64/vmware.base.json
```diff
@@ -0,0 +1,76 @@
+{
```
ekinanp: Imaging pipeline wise, this might mean that we'd need to rebuild the Mac OS 10.13 template whenever Mac OS 10.12 changes. @jcoconnor What do you think?

templates/macos/common/vmware.vsphere.nocm.json
```diff
@@ -0,0 +1,91 @@
+{
+  "description"                             : "Builds a vSphere template using an existing base macOS template build",
+  "_comment"                                : "This is a common vmware vsphere template for macOS platforms; It customizes a macOS 10.12 or 10.13 base image for use with vmpooler. Null variables should be provided by vars.json files within the version-specific subdirectories, or at the command line",
+  "variables": {
+    "template_name"                         : null,
+    "beakerhost"                            : null,
+    "version"                               : null,
```
ekinanp: Would be helpful to state all required variables here -- including things like "headless" (which you can default to true and thus take it out of the common/vars file).

templates/macos/10.12/x86_64/vmware.base.json
caseywilliams: What "current build script" is this? I don't think I've seen it?

ekinanp: https://gist.github.com/ekinanp/2d63e8a6339ff7ecb9d3e8002cc5a172

I think I can modify https://gist.github.com/ekinanp/2d63e8a6339ff7ecb9d3e8002cc5a172#file-packer_build-sh-L55 to add a check after to see if there's a var file in the template directory and add that as part of the var-files and things should work.

caseywilliams: Why not put that script in this repo or link to it in the README? Seems important.

ekinanp: I talked about it with @ScottGarman a few weeks back. We agreed that the relevant README should contain the build instructions for each template so that there is no need for people to use the script.

templates/macos/10.13/x86_64/vmware.base.json
caseywilliams: It would definitely mean that - I can't see the 10.12 image changing with any regularity though.

templates/macos/10.13/x86_64/vmware.base.json
caseywilliams: The state of xcode remains the same in the 10.13 image after the upgrade, but I'm glad you asked about this, because they both seem to still need a run of `xcode-select` somewhere in the setup scripts - I'll add that and verify today.

README.md
```diff
@@ -8,30 +8,38 @@ The Packer templates are organized under the following directory structure insid
 ```
 <os-family>/<os-distribution>/<os-variant>/<architecture>
 ```
-where examples of `<os-family>` are Linux, Windows; of `<os-distribution>` are RedHat, Ubuntu; of `<os-variant>` are 6 and 7 with RedHat as the `<os-distribution>`; and of `<architecture>` are `x86_64`, `i386`.
+
+1. `<os-family>` is the OS family, i.e. `linux`, `windows`, or `macos`
```
ekinanp: Might not be a bad idea to have a `solaris` subdirectory too. @ScottGarman What do you think?

templates/macos/10.13/x86_64/vmware.base.json
caseywilliams: Oh, I just had `install_xcode_cli_tools` set to false by default 🙄

templates/macos/10.13/x86_64/vmware.base.json
```diff
@@ -0,0 +1,76 @@
...
+    "installer_url"                                : null
+  },
+
+
+  "builders"                                       : [
+    {
+      "name"                                       : "{{user `template_name`}}-{{user `provisioner`}}-base",
+      "type"                                       : "vmware-vmx",
+      "vm_name"                                    : "packer-{{build_name}}",
+      "source_path"                                : "{{user `source_directory`}}/macos-1012-x86_64-{{user `provisioner`}}-base/macos-1012-x86_64-{{user `provisioner`}}-base.vmx",
```
ekinanp: Might be good to have "macos-1012-x86_64" be a user variable since it's used twice here.

README.md
ScottGarman: Agree here on the solaris subdir

ekinanp: Ok, will update https://github.com/puppetlabs/puppetlabs-packer/pull/249 once this gets merged in.

templates/macos/10.13/x86_64/vmware.base.json
geoffnichols: fwiw, it looks like the 10.12 image currently in the pooler is a Sierra beta:

```
# sw_vers
ProductName:	Mac OS X
ProductVersion:	10.12
BuildVersion:	16A312a
```

caseywilliams: It'd be great if we could find a newer 10.12 installer package that was still from pre-10.12.4, when the pkg signing requirement was introduced, but at this point unless we can find one, it's unlikely we'll be able to get a new 10.12 image unless we find a way to get the installer to trust the pkg.

For what it's worth, if you try to include a customization pkg in a 10.12.4 image without signing it, the install process aborts with a "Package is not signed" message. I tried signing the pkg with `Mac Installer Distribution` and `Developer ID Installer` developer certs, but in both cases the installer aborted with "Package is not trusted" instead.

Looking at the pkgs that are built into the 10.12.4 installer's ESD image (amongst which we used to put the customization pkg), I see that they're signed with the following:

```
   Status: signed Apple Software
   Certificate Chain:
    1. Software Update
       SHA1 fingerprint: ...
       -----------------------------------------------------------------------------
    2. Apple Software Update Certification Authority
       SHA1 fingerprint: ...
       -----------------------------------------------------------------------------
    3. Apple Root CA
       SHA1 fingerprint: ...
```

My own signed packages have a different intermediary CA (`Developer ID Certification Authority` or `Apple Worldwide Developer Relations Certification Authority`). Perhaps there is some other way to sign them? I'd love it if we could find a way for the installer to trust the package but I haven't been able to figure it out - Apples doc's are frustratingly opaque about this.

templates/macos/common/vars.json
jcoconnor: Do we need a common ```.vars.json``` file - would prefer to put the defaults in the generic (base or vmware/cygwin stage).

templates/macos/common/vmware.vsphere.nocm.json
jcoconnor: Thats correct.

caseywilliams:
Putting a don't merge label on this until I update it to include an APFS conversion option

caseywilliams:
Updated with APFS support

ScottGarman:
This is at a good checkpoint to merge and make further improvements as needed.

jcoconnor:
Just saw merge go thru as I was about to give a 👍 looks good overall, but specific details probably need tested. Like the idea of moving OSX scripts into the common OSX area and am going to copy this in windows.

PR #258, opened 2018-01-22
(maint) Update vsphere setup script to work with official RedHat images

ScottGarman:
This was a tweak I had to make to get the redhat fips image to build. Previously we haven't had any official RedHat templates in this repo, only CentOS/Scientific/Oracle.


McdonaldSeanp:
👍

PR #259, opened 2018-01-25
(IMAGES-721) Restructure Windows Directories

jcoconnor:
Moving away from the nested OS group model under windows to a top-level
"win" directory with the common directory under it.

This will require some adjustment of the common .json files to pick up
the new location.


jcoconnor:
I've tested this locally and it looks good enough to merge.

PR #260, opened 2018-01-25
(MAINT) Flatten the templates/linux directory + update READMEs

ekinanp:


ekinanp:
Verified locally that the build-script will work on Debian (one w/ a separate base template), Redhat FIPS (separate scripts), and Ubuntu (representative Linux dist. that doesn't have any special customizations) for vmware.base.json.

templates/common/vmware.base.json
```diff
@@ -11,7 +11,7 @@
     "floppy_files"                                 : null,
     "boot_command"                                 : null,
 
-    "project_root"                                 : "../../../../..",
+    "project_root"                                 : "../../../..",
```
jcoconnor: Will take this point up separately from the PR review, but I think we should probably use a Root Point variable for this in both Windows and Linux json files.

PR #261, opened 2018-01-29
Update to replace buildsources with Artifactory

shrug:


shrug:
@jcoconnor It looks like the other templates were updated, either this one was overlooked or very possibly it's not being built anymore. Here's an update for good measure :)

jcoconnor:
Thanks @shrug that looks 👍 Will merge it now.

PR #264, opened 2018-02-09
(WIN-73) Initial Win-2008r2 GCE Template

jcoconnor:
First pass at Windows 2008r2 GCE Template.
This is being added as a stub for the moment, as there are some issues
with this method that need to be resolved.

In particular, unable to translate variables in the powershell provider
which has been opened as a packer issue (5878)

There is also the question if we can obtain the winrm_password which is
set using the GCE Api - this will be needed to enable auto-login which
is required for Appveyor

Submitting this in the interim as practice code.


jcoconnor:
For reference, the Packer issue logged is at: https://github.com/hashicorp/packer/issues/5878

PR #270, opened 2018-02-25
RHEL 6.8 and 7.2 Templates and Improvements

ScottGarman:
I think somewhere down the line, the ethernet0 default of vmxnet3 disappeared from the vsphere template. This restores it but in a manner that allows for variable overrides if for some reason a platform doesn't want to use that device (we generally want it used everywhere possible for better networking performance).

It also eliminates the need for using the enable-rhel-repos bootstrap script and puppetizes the configuring of a default yum mirror for RHEL vmware base template installs.I'll finally remove that script after updating the RHEL 7.0-fips template.

### DIFF
--- a/manifests/modules/packer/manifests/repos.pp
+++ b/manifests/modules/packer/manifests/repos.pp
@@ -1,0 +1,37 @@
+# For platforms that come up without any configured software repos
+# (e.g, RHEL), this class is used to configure repos pointing to
+# our internal mirrors so that packages can be installed by other
+# puppet classes (mainly vmtools.pp).
+#
+# TODO: Consolidate this and the vsphere repos manifest into a single
+# module, which can be conditionally included in base.pp for targeted
+# platforms and for all cases in vsphere.pp.
+class packer::repos {
+
+  if $::operatingsystem == 'RedHat' {
+
+    $repo_mirror = 'https://artifactory.delivery.puppetlabs.net/artifactory'
+    $os_mirror   = 'http://osmirror.delivery.puppetlabs.net'
+    $gpgkey      = 'RPM-GPG-KEY-redhat-release'
+
+    resources { 'yumrepo':
+      purge => true,
+    }
+
+    # We don't have consistent mirror urls between RedHat versions:
+    $base_url = $::operatingsystemmajrelease ? {
+      '7' => "${repo_mirror}/rpm-rhel/7.2/${::architecture}",
+      '6' => "${repo_mirror}/rpm-rhel/6.8/${::architecture}",
+      '5' => "${os_mirror}/rhel50server-${::architecture}/RPMS.all"
+    }
+
+    yumrepo { "localmirror-everything":
+      descr    => "localmirror-everything",
+      baseurl  => "${base_url}",
+      gpgcheck => "1",
+      gpgkey   => "file:///etc/pki/rpm-gpg/${gpgkey}"
+    }
+
+  }
+
+}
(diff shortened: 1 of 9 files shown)


ScottGarman:
The PR to enable these platforms in our imaging pipelines is: https://github.com/puppetlabs/ci-job-configs/pull/4087

manifests/modules/packer/manifests/vsphere/repos.pp
```diff
@@ -77,11 +77,19 @@
...
-        'OracleLinux' => "${os_mirror}/${loweros}-${::operatingsystemmajrelease}-${::architecture}/RPMS.all" 
+      if $::operatingsystem == 'RedHat' {
+        # We don't have consistent mirror urls between RedHat versions:
+        $base_url = $::operatingsystemmajrelease ? {
+          '7' => "${repo_mirror}/rpm-rhel/7.2/${::architecture}",
+          '6' => "${repo_mirror}/rpm-rhel/6.8/${::architecture}",
+          '5' => "${os_mirror}/rhel50server-${::architecture}/RPMS.all"
+        }
+      } else {
+        $base_url = $::operatingsystem ? {
```
ekinanp: You could probably do a nested selector here, something like:

```
$base_url = $::operatingsystem ? {
  ...
  RedHat => $::operatingsystemmajrelease ? {
    '7' => ...
    ...
  }
}
```

That would avoid the extra if statement. Although at that point, it might be worthwhile to use a Case statement?

manifests/modules/packer/manifests/repos.pp
```diff
@@ -0,0 +1,33 @@
+# For platforms that come up without any configured software repos
```
ekinanp: I wonder if it would be useful to remove the vsphere/repos.pp manifest into its own module (i.e. create repos/params.pp file and then repos.pp file) so all of the repo configuration logic would be in one place. We can then conditionally include that in the base.pp file for RHEL operating systems and then always include it in the old vsphere.pp file.

It does seem weird that we have the repo manifest under vsphere, since that looks like something which could be used in other places.

That can be work done later though, given the immediate necessity of these changes. Although it does not like something that's too difficult to do.

manifests/modules/packer/manifests/repos.pp
```diff
@@ -0,0 +1,33 @@
...
+# (e.g, RHEL), this class is used to configure repos pointing to
+# our internal mirrors so that packages can be installed by other
+# puppet classes (mainly vmtools.pp).
+class packer::repos {
+
+  if $::osfamily == 'RedHat' and $::operatingsystem == 'RedHat' {
```
ekinanp: I think you can just use $::operatingsystem == 'RedHat' here?

manifests/modules/packer/manifests/vsphere/repos.pp
ScottGarman: A case statement might make more sense. I'm hoping though soon that RelEng will be able to mirror the RHEL 5 repos on artifactory using the same url scheme as 6 & 7 and we can consolidate all of those cases into one. They'll need some more time for that though.

manifests/modules/packer/manifests/repos.pp
ScottGarman: Agreed. I've added a TODO comment to this effect to return to it later.

manifests/modules/packer/manifests/repos.pp
ScottGarman: Good catch, updated.

ScottGarman:
Merging due to the urgency of the CI blocking nature of the work.

PR #271, opened 2018-02-28
bump to 1.2.1

witjoh:
0.12.0 is really old, so bumped to the latest version.

Also started adding vagrant configs for centos 7.4 for
* vagrant-libvirt
* vagrant-virtualbox
* vagrant docker
both for nocm and puppet configs.
Don't know what the policies will be from your side .... Should I create a PR or just keep it in a separate project ???

### DIFF
--- a/tests/test-setup.sh
+++ b/tests/test-setup.sh
@@ -10,7 +10,7 @@ if [ ! -f packer ] ; then
         arch="amd64"
     fi
     os=$(uname |tr '[:upper:]' '[:lower:]')
-    version="0.12.0"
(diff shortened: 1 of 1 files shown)


shrug:
Ping @ScottGarman - I think it's worth opening PRs for vagrant configs, but I'll let @ScottGarman verify

ScottGarman:
@witjoh If the vagrant templates follow the new hierarchical var file format we've started using, I would take a close look at accepting that here, so please do open a PR for them. I'm :+1: on this PR.

PR #274, opened 2018-03-05
(maint) Update redhat-fips 7.2 image to use new mirror repos

ScottGarman:
This just amounts to a version bump, as a rebuild will pick up the new
repo scheme as defined in repos.pp.


ScottGarman:
I've built and tested this locally to verify the new repo urls are being used and that package groups are properly working.

ScottGarman:
Merging as the change is trivial and the fix is urgent for CI.

PR #276, opened 2018-03-07
Centos 7.4

witjoh:
@ScottGarman 
Except for the little modification in the puppet code, everything is separated form the other base code.
I think the naming of the vars I used are subject to changes to follow the naming convention ...

Adds vagrant boxes nocm and puppet for vagrant supporting following hypervisors : 
* libvirt (kvm)
* virtualbox
* docker

### DIFF
--- a/manifests/modules/packer/manifests/networking/params.pp
+++ b/manifests/modules/packer/manifests/networking/params.pp
@@ -8,7 +8,7 @@ class packer::networking::params {
 
     redhat: {
       case $::operatingsystemrelease {
-        '7.0', '7.0.1406', '7.1.1503', '7.2.1511', '7.2', '7.3.1611': {
+        '7.0', '7.0.1406', '7.1.1503', '7.2.1511', '7.2', '7.3.1611', '7.4.1708': {
           case $::provisioner {
             'virtualbox': { $interface_script = '/etc/sysconfig/network-scripts/ifcfg-enp0s3' }
             'vmware':     { $interface_script = '/etc/sysconfig/network-scripts/ifcfg-ens33' }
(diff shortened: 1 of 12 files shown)


ScottGarman:
Thanks for this @witjoh. I'm open to incorporating this work in our repo, provided we clearly document which templates are officially maintained vs. community maintained (which we can take care of in a separate PR). My main concern is that we can do our work without stepping on each other, which appears to be the case so far.

@ekinanp @jcoconnor what are your thoughts?

ekinanp:
@ScottGarman I think explicitly documenting the official vs. community templates is a good idea -- this would probably go in the top-level README? Otherwise I don't see an issue here since, like you pointed out, our work vs. what's here is independent of each other.

witjoh:
As for as I see, the only files we will be editing both are the most specific distro ones.  And by having a clear naming convention for the used variables, we sure can keep everything separated properly.
Could be in tine there will be some modification needed in the used puppet code, but they will be in most cases very small.  Documenting the template rules is indeed something we need.  I think community will be mostly interested in the vagrant based templates, supporting different hypervisors, which the ones included in this PR (kvm/virtualbox/docker) will be the most interested ones. 
One ting that I should change in the templates are the URI of the distro's iso files.  As you see I used here a public available one.

mradcliffe:
Yes, I'm mostly interested in the vagrant-based templates, and it's good to see support again :-)

The CentOS 7.4 template and vars look pretty much the same to what I was starting to maintain on in my own [fork](https://github.com/mradcliffe/puppetlabs-packer/blob/feature/selinux-vars/templates/centos/7.4/x86_64/vars.json) so that looks good.

witjoh:
Thx for merging

PR #277, opened 2018-03-07
(MAINT) Update Version on all Windows Builds

jcoconnor:
Set the version to todays date to ensure no clashes with previous
builds.


jcoconnor:
@ScottGarman or @ekinanp Can I get a merge on this.
Just updating the version string for all the windows builds so that I don't clash with previous builds.
This is in preparation for the Patch Tuesday update which I will be starting next week.

ScottGarman:
Merging, as this is a trivial change.

PR #285, opened 2018-03-14
(IMAGES-783) Disable firewall service for redhat-6 templates

ScottGarman:
During the update from RHEL 6.5 -> 6.8, RedHat enabled a firewall
service by default. We disable this in the redhat-7 template build
process, so do this also for redhat-6.


ScottGarman:
These templates have been validated, passed imaging CI, and will be going into production today.

PR #287, opened 2018-03-19
(maint) CentOS 6 updates

ScottGarman:
This includes some minor tweaks to the CentOS 6 templates:

* Updated x86_64 template to have 6 GB of RAM
* Updated puppet-agent version to 1.10.10
* Updated README file to explain why we have separate build instructions
for different versions of CentOS


ScottGarman:
These templates build locally and I've verified our mirror repos are working with yum package groups, the firewall is disabled, etc.

PR #292, opened 2018-03-20
enable redhat repo url config using env vars

witjoh:
Attempt to make the uri settings for redhat mirrored repos configurable using environment.  This introduces following envs : 
PACKER_REDHAT_OS_REPO = someurl
PACKER_REDHAT_OPTIONAL_REPO = someurl
PACKER_REDHAT_EXTRA_REPO = someurl

These are translates to facts in the *.base.json.
Whenever unset or all are empty, we include the stock packer::repo class, otherwise we include the packer::community_repso class.  Last one include only those repos if the environment is set.

### DIFF
--- a/manifests/base.pp
+++ b/manifests/base.pp
@@ -3,6 +3,6 @@ include packer::updates
 include packer::sshd
 include packer::networking
 
-unless $::provisioner == 'ec2' or $::provisioner == 'libvirt' {
+unless $::provisioner in [ 'ec2', 'libvirt', 'virtualbox', 'docker' ] {
   include packer::vmtools
 }
--- a/manifests/modules/packer/manifests/repos.pp
+++ b/manifests/modules/packer/manifests/repos.pp
@@ -1,51 +1,12 @@
-# For platforms that come up without any configured software repos
-# (e.g, RHEL), this class is used to configure repos pointing to
-# our internal mirrors so that packages can be installed by other
-# puppet classes (mainly vmtools.pp).
-#
-# TODO: Consolidate this and the vsphere repos manifest into a single
-# module, which can be conditionally included in base.pp for targeted
-# platforms and for all cases in vsphere.pp.
+# warapper class to select which repos to install
+# this way we can run this class on ots own - needed by docker
 class packer::repos {
 
-  if $::operatingsystem == 'RedHat' {
-
-    $repo_mirror = 'https://artifactory.delivery.puppetlabs.net/artifactory'
-    $os_mirror   = 'http://osmirror.delivery.puppetlabs.net'
-    $gpgkey      = 'RPM-GPG-KEY-redhat-release'
-
-    resources { 'yumrepo':
-      purge => true,
-    }
-
-    # We don't have consistent mirror urls between RedHat versions:
-    # TODO: RHEL 5 needs further refactoring
-    $base_url = $::operatingsystemmajrelease ? {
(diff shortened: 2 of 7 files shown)


witjoh:
I had to rearrange the modules a bit, to be able to build redhat based docker vagrant boxes. 
I needed to run the class packer::repos on such to be able to install the redhat local repos. 
On docker images the base module is not usable.  I believe this way we can also keep the  some community stuff separated from your stuff

ScottGarman:
Hi @witjoh, apologies but I have had to complicate this a bit with several changes to the yum repo configs for CentOS and Scientific Linux in https://github.com/puppetlabs/puppetlabs-packer/commit/8e71526960412338e5b3a5064c485ff3ce88e3ec

I agree that it would be best to make the repo urls be configurable via environment vars in packer as you're intending to do here. Unfortunately I've found that between the different EL versions, the names and numbers of repositories used tend to differ in inconvenient ways (e.g, updates vs. extras vs. fastbugs vs. security, etc). Also the way we have our internal artifactory mirrors set up doesn't necessarily follow the same format as public mirror urls.

For the sake of simplicity, I'm wondering if it might be most straightforward to set a single environment variable (e.g, external_mirrors) that could be used in the puppet code to branch between internal vs. external mirror repo configurations. That way we don't need to go about naming new variables for each OS and repo combination. Thoughts? cc @ekinanp

witjoh:
That's fine by me. The only thing I need is to be able to classify the class handling the repo code on its own, since I need to do that for vagrant-docker base image, especially for enterprise distros, since they are not registred and need local mirrors.  We could just provide a long list of uris pointing to some repo, and just generate whatever is needed.  Community distros don't have this issue, but could indeed use the same mechanism to point to internal repos. I will refactor the code and update this PR.

ekinanp:
@ScottGarman @witjoh If I am understanding correctly, the issue is that we want to find a clean way to let people use external repo mirrors to configure their images, right? If so, I agree with @ScottGarman on using a single environment variable. I think the environment variable can be something like an enum, like `REPO_MIRROR`, that can have different values (e.g. `internal`, `some_mirror_one`, `some_mirror_two`, etc..). The puppet repo code itself can branch out to the appropriate class (e.g. 'internal' => repos/puppetlabs, 'some_mirror_one' => repos/mirror_one, 'some_mirror_two' => repos/mirror_two, etc.) based on this variable's value.

witjoh:
@ekinanp @ScottGarman 
Looking at your proposal, it seems we need to create custom repos classes for every value we want to be able to assign to the REPO_MIRROR env ?  Which in my opinion looses flexibility, since we only need to be able to override the 'baseurl' setting in the yum configs ... well for redhat at least ...

ekinanp:
@witjoh @ScottGarman Yeah ignore me, I didn't know at the time of writing that that you needed to specify three yumrepo resources (os, optional, extras) when installing a RHEL image itself; I thought it was something weird that we did internally because that also happens a lot. I'm in agreement w/ @ScottGarman.

mwaggett:
What is the status of this? At the very least, needs a rebase.

ScottGarman:
This PR has mostly become a design RFC and isn't actionable at this time. It could use another attempt but I'm not sure when I'm going to be able to get to this in the near future.

mwaggett:
Since this hasn't seen any activity in several years, I'm going to go ahead and close this pull request. Please re-open this if you plan to resume work on it.

If you have any questions or concerns, please don't hesitate to ping us in slack.puppet.com.

PR #297, opened 2018-04-05
(IMAGES-752) Ubuntu 18.04 systemd updates

caseywilliams:
18.04 commits more heavily to systemd than the previous LTS; Update
several parts of the vsphere bootstrap script for 18.04 to rely on
systemd tools:

- Run the bootstrap script via a oneshot vsphere.bootstrap systemd
  service instead of via rc.local
- systemd-networkd's DHCP client uses /etc/machine-id to create a client
  ID - this value is the same on hosts cloned from the same template, so
  regenerate it during bootstrap
- Use systemd-timesyncd instead of NTP
- Use hostnamectl to set the hostname

### DIFF
--- a/templates/ubuntu/18.04/common/files/preseed.cfg
+++ b/templates/ubuntu/18.04/common/files/preseed.cfg
@@ -21,11 +21,12 @@ d-i passwd/root-password password puppet
 d-i passwd/root-password-again password puppet
 d-i passwd/make-user boolean false
 d-i user-setup/allow-password-weak boolean true
-d-i pkgsel/include string curl gcc make nfs-common ntp openssh-server perl wget
+d-i pkgsel/include string curl gcc make nfs-common openssh-server perl wget
 d-i pkgsel/install-language-support boolean false
 d-i pkgsel/update-policy select none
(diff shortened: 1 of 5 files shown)


templates/ubuntu/18.04/common/files/preseed.cfg
```diff
@@ -21,11 +21,12 @@ d-i passwd/root-password password puppet
...
 d-i passwd/make-user boolean false
 d-i user-setup/allow-password-weak boolean true
-d-i pkgsel/include string curl gcc make nfs-common ntp openssh-server perl wget
+d-i pkgsel/include string curl gcc make nfs-common openssh-server perl wget
 d-i pkgsel/install-language-support boolean false
 d-i pkgsel/update-policy select none
 d-i pkgsel/upgrade select full-upgrade
-d-i time/zone string US/Pacific
+d-i time/zone string GMT
+d-i clock-setup/utc boolean true
```
caseywilliams: IMAGES-802 has been filed to use UTC on Ubuntu images instead of PST (all other platforms use UTC)

PR #300, opened 2018-04-12
(IMAGES-801) Add the load balancer repo to redhat-6-x86_64 images

ScottGarman:
Puppet modules testing CI requires the haproxy package, which is offered
via the lb (load balancer) repo. This is only required for the
redhat-6-x86_64 platform.


ScottGarman:
I have successfully validated this image on vmpooler-dev. The Jenkins imaging acceptance tests failed for the same reason that p-a 5.5.x puppet tests are failing. So that's a known issue and the change being made to the image here would not cause the test failure. This is ready for merge now.

PR #301, opened 2018-04-13
(IMAGES-806) Update Win-2016-hyperv to Build 1709

jcoconnor:
Hyperv/Docker requires Build 1709 which is a seperate ISO from the
original updated windows build .iso. 

Note - This is the CORE rather than the full Desktop/GUI edition of windows.

The first commit is the actual change, the second commit is just an update of the version strings across all the builds as is current practice until we automate versioning for the windows builds.


michaeltlombardi:
Travis error looks unrelated but still needs resolving.

ScottGarman:
According to our travis config, it looks like we're running this test on Ubuntu Trusty, but are installing the shellcheck package from the Debian sid repository. We should look into updating this to use Ubuntu Xenial, which offers shellcheck in its native repos.

PR #302, opened 2018-04-23
(IMAGES-747) Add Fedora 28 (beta) platform

ScottGarman:


ScottGarman:
This has been tested/validated on vmpooler-dev successfully.

templates/fedora/28/x86_64/vars.json
```diff
@@ -0,0 +1,13 @@
...
+    "template_name"                         : "fedora-28-x86_64",
+    "template_os"                           : "fedora-64",
+    "beakerhost"                            : "fedora28-64",
+    "version"                               : "0.0.1",
+    "iso_url"                               : "https://artifactory.delivery.puppetlabs.net/artifactory/generic/iso/Fedora-Server-dvd-x86_64-28_Beta-1.3.iso",
+    "iso_checksum"                          : "73efad9c6f17055d3166042eebf9946ff8f16071fbb1f4a53cc9ded8771bb068",
+    "iso_checksum_type"                     : "sha256",
+    "vmware_vsphere_nocm_vmx_data_memsize"  : "4096",
+    "vmware_vsphere_nocm_vmx_data_numvcpus" : "2",
+    "floppy_files"                          : "../files/ks.cfg",
```
ekinanp: Would be more consistent to have this in a 'common/files' folder but not a big deal.

ScottGarman: If we were building multiple architectures for F28, I can see that common would have some semantic benefits. But in this case the kickstart file is specific only to this arch of F28.

PR #306, opened 2018-04-30
Documentation improvements re: support status

ScottGarman:
Adds a support_status variable to all templates to indicate whether the template is maintained by Puppet vs. community maintained. It also adds a Support Status section in the various OS README.md files to clarify which OS versions are Puppet maintained vs. community maintained.

### DIFF
--- a/templates/centos/README.md
+++ b/templates/centos/README.md
@@ -2,8 +2,13 @@
 
 ### About
 
-This contains all of the var files required to build any of the templates in the
-`templates/common` directory for Centos platforms.
+This contains all of the var files required to build any of the templates in the `templates/common` directory for Centos platforms.
+
+### Support Status
+
+This repository includes OS platforms that are officially supported at Puppet and ones that are entirely maintained by the community. Packer templates include a `support_status` variable which indicates whether the template is puppet maintained vs. community maintained.
+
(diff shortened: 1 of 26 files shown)


templates/centos/README.md
```diff
@@ -2,8 +2,13 @@
 
 ### About
 
-This contains all of the var files required to build any of the templates in the
-`templates/common` directory for Centos platforms.
+This contains all of the var files required to build any of the templates in the `templates/common` directory for Centos platforms.
+
+### Support Status
+
+This repository includes OS platforms that are officially supported at Puppet and ones that are entirely maintained by the community. Packer templates under the `templates/common` directory include a `support_status` variable which indicates whether the template is puppet maintained vs. community maintained.
```
ekinanp: Might be enough to say "Packer templates include" instead of "Packer templates under the 'templates/common' directory"? Because we can have templates that are not in the `templates/common` directory (e.g. for the VRO and Debian images).

ScottGarman: Ah good point - I forgot to add the support_status var to those templates, too. Will fix shortly.

ScottGarman:
PR updated based on feedback from @ekinanp

PR #307, opened 2018-05-03
(IMAGES-752) Update Ubuntu 18.04 image to final GA version

ScottGarman:
This updates the image from the beta2 release we were using previously.


jcoconnor:
👍

PR #313, opened 2018-05-11
(maint) Fix OSX base templates to work with (future) CI

ekinanp:
Also makes them a bit more consistent with the existing templates in
templates/common


ScottGarman:
:+1:

PR #319, opened 2018-05-14
(IMAGES-822) Fix Network Adapter error for Win-2008R2 WMF

jcoconnor:
Need to add a special case for Win-2008R2 WMF5 as although the Get-NetAdapter cmdlet may be present, it is not supported, so filter it out.


jcoconnor:
The remaining vmpooler image ```win-2008r2-wmf5-x86_64``` is now initialising correctly and talking nicely to the test suite.

PR #322, opened 2018-05-17
(IMAGES-825) Set NOACL for all cygwin file creates

jcoconnor:
Setting the NOACL flag in the cygwin /etc/fstab to ensure ACL is not
added to any file creates

### DIFF
--- a/templates/win/10-1511/x86_64/vars.json
+++ b/templates/win/10-1511/x86_64/vars.json
@@ -6,7 +6,7 @@
     "windows_version"   : "Windows-10",
     "image_name"        : "Windows 10 Enterprise 2015 LTSB",
     "product_key"       : "WNMTR-4C88C-JK8YV-HQ7T2-76DF9",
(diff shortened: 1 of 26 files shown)


jcoconnor:
Since this is changing cygwin default file properties I need to put it through decent acceptance testing before even thinking of merging this change.

It does appear to fix the issue when you configure it manually on a vmpooler machine though.

Iristyle:
Based on my understanding, this looks good to me.

My only concern is whether or not we have any Beaker tests around file modes that look at the "Cygwin view" of the filesystem.  That is, shell commands embedded in tests, that parse the output of `ls`, `stat` or similar. I suspect the answer is no, and that we rely on Puppets internal mode calculation in such instances, which wouldn't be impacted by Cygwin.  We do have a number of "integration" tests that run on Windows (but outside of Beaker), and those use APIs directly, so I don't believe they are impacted.

👍

PR #323, opened 2018-05-18
(IMAGES-576) Use powershell latest module for puppet configure

jcoconnor:
This change reverts compatibility issue with earlier versions of the Powershell module
where had to pin to version 1.0.6 and use some compatibility code for the gpedit
functions.

Since the issues are no longer a problem, remove the pin and compatibility code.


jcoconnor:
This has been tested on the following windows platforms and executes the puppet configure stage without issues:
- win-2008
- win-2008r2
- win-2008r2 wmf5
- win-2012r2
- win-10-pro

PR #324, opened 2018-05-18
(IMAGES-824) Bump the template version number in vars.json in centos 7.2

8675309:
(IMAGES-824) Bump the template version number in vars.json in centos 7.2 to read from the new template that adds the extras repository, allowing installation from Docker CE.


ScottGarman:
:+1:

PR #325, opened 2018-05-24
(IMAGES-834) Use env. var. to specify raw VMX file for VRO images

ekinanp:
The original template specified the source path for the VRO vmx file
using the same convention for vmware.base images. This was not the right
way to think about building VRO images, because we do not build a base
image. Instead, we download the raw OVA and convert it to its raw VMX
format, slightly modifying its networking config. to use shared instead
of bridged networking.

This commit modifies the vmware.vsphere.nocm template to take in the
path to the raw VMX file via. an environment variable instead, which
also simplifies the CI code for our VRO images.


ekinanp:
This is blocked on https://github.com/puppetlabs/platform-ci-utils/pull/12.

ekinanp:
Unblocked now.

ScottGarman:
:+1:

PR #327, opened 2018-05-25
(WIP) IMAGES-734 SPIKE - Evaluate vsphere-iso packer plugin

jcoconnor:
This PR is for Proof of Concept/Spike purposes only.
It is based on the [vsphere-iso plugin](https://github.com/jetbrains-infra/packer-builder-vsphere) plugin which offloads the actual image building onto a vSphere server.

This POC demonstrates the plug-in for the ```win-10-ent-x86_64``` & ```win-10-ent-x86_64``` images only.

### DIFF
--- a/templates/win/common/files/AutoUnattendTemplate.xml
+++ b/templates/win/common/files/AutoUnattendTemplate.xml
@@ -9,6 +9,14 @@
 -->
 <unattend xmlns="urn:schemas-microsoft-com:unattend">
     <settings pass="windowsPE">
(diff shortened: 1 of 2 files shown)


ekinanp:
Yup this makes sense. Per our discussion earlier, one nice feature about this template is it could lead us to containerize Jenkins instances in the imaging pipelines (because the hard work of building the image is done on VCenter). Plus, it also gives us an easy way to access a VM's console directly for debugging instead of having to set-up VNC and connecting with it to the VM on one of the builders.

My only concern is a lot of built templates on VCenter that we'd have to manage (e.g. to delete a template we no longer need, for example). The UI for that already sucks enough. But with the new ci-utils gem, we could add some basic functionality for doing this (or evangelize an existing tool https://github.com/mattkirby/py-vmware).

It might be worthwhile to do a SPIKE at some point in the future on converting the Linux templates to this.

jcoconnor:
Closing this PR in favour of #344

PR #328, opened 2018-05-25
(IMAGES-841) Remove overwrite option from vmware.vsphere.nocm.json

ekinanp:


templates/common/vmware.vsphere.nocm.json
```diff
@@ -126,8 +126,7 @@
       "vm_folder"                           : "{{user `packer_vcenter_folder`}}",
       "vm_name"                             : "{{user `template_name`}}-{{user `version`}}",
       "vm_network"                          : "{{user `packer_vcenter_net`}}",
-      "insecure"                            : "{{user `packer_vcenter_insecure`}}",
-      "overwrite"                           : "true"
```
jcoconnor: Could you explicitly put this in as "false" rather than relying on it as a default.

templates/common/vmware.vsphere.nocm.json
```diff
@@ -140,8 +139,7 @@
       "vm_folder"                           : "{{user `packer_vcenter2_folder`}}",
       "vm_name"                             : "{{user `template_name`}}-{{user `version`}}",
       "vm_network"                          : "{{user `packer_vcenter2_net`}}",
-      "insecure"                            : "{{user `packer_vcenter2_insecure`}}",
-      "overwrite"                           : "true"
```
jcoconnor: As above

PR #329, opened 2018-05-28
(IMAGES-31) Pin Powershell and other apps to the taskbar

jcoconnor:
This PR requires two different methods depending on platform:
1. Windows-10/2016 uses the new StartLayout GPO policy to define taskbar items.
2. Windows-2012r2 and earlier uses a simple script run during the post-clone configuration.

The <TaskBarLinks> elements used in the Post-Clone Autounattend have also been removed as they didn't work reliably.

### DIFF
--- a/templates/win/common/scripts/vmpooler/vmpooler-post-clone-configuration.ps1
+++ b/templates/win/common/scripts/vmpooler/vmpooler-post-clone-configuration.ps1
@@ -125,6 +125,17 @@ Set-Service "sshd" -StartupType Automatic
 schtasks /create /tn UpdateBGInfo /ru "$AdministratorName" /RP "$qa_root_passwd" /F /SC Minute /mo 20 /IT /TR 'C:\Packer\Scripts\sched-bginfo.vbs'
 schtasks /run /tn UpdateBGInfo
 
+# Pin apps to taskbar as long as we aren't win-10/2016
+
+if ($WindowsVersioni -notlike $WindowsServer2016) {
+  try {
+    Write-Output "Pin Apps to Taskbar"
+    & $PackerScripts\Pin-AppsToTaskBar.ps1
+  }
+  catch {
+    Write-Output "Ignoring Pin App errors"
+  }
+}
 # Rename this machine to that of the VM name in vSphere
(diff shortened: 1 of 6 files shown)


jcoconnor:
I've tested this right across all the builds for vSphere/Packer - the taskbar icons are being displayed across all and acceptance tests run ok.

jcoconnor:
And I'm going to leave Explorer to another day, or not at all - having Powershell sorted is after all the title of the PR.

templates/win/common/scripts/vmpooler/vmpooler-post-clone-configuration.ps1
```diff
@@ -125,6 +125,17 @@ Set-Service "sshd" -StartupType Automatic
 schtasks /create /tn UpdateBGInfo /ru "$AdministratorName" /RP "$qa_root_passwd" /F /SC Minute /mo 20 /IT /TR 'C:\Packer\Scripts\sched-bginfo.vbs'
 schtasks /run /tn UpdateBGInfo
 
+# Pin apps to taskbar as long as we aren't win-10/2016
+
+if ($WindowsVersioni -notlike $WindowsServer2016) {
```
michaeltlombardi: Will this still work with the upcoming Windows Server 2019 release?  If not it may not need to be addressed here but should be investigated if we plan to support that platform in the runup to its release.

jcoconnor: Good point and I checked it.
Windows-2019 preview release still returns the Windows-2016 kernel version.
Not sure if thats intentional, but will follow it as we track pre-releases.

The layout looks good, an in any case, even if the script fails, the try/catch will ignore the failure and continue to the end of the initialisation.

michaeltlombardi: Good enough for me, we can address this in a future 2019-proofing ticket anyway.

michaeltlombardi: 🤦‍♂️

PR #333, opened 2018-06-07
(IMAGES-855) Add libtool to the cygwin packages

jcoconnor:
Libtool is installed/configured by libcurl and appears to be having
issues, so try installing this directly from cygwin to see if it
performs better.

The cygwin setup.exe is also well out of date, so grab this from cygwin
directly instead of our own artifactory.


jcoconnor:
Abandoning this work as no longer needed - separate PR coming up to revert to previous ```2.6.0``` version of Cygwin

PR #341, opened 2018-06-21
(IMAGES-852) Fix BGinfo Error and other small packer/windows issues

jcoconnor:
Small fixes to the Windows/Packer builds including:
1. IMAGES-852 - Fix initial BGInfo run error

### DIFF
--- a/templates/win/common/puppet/windows_template/manifests/registry/machine.pp
+++ b/templates/win/common/puppet/windows_template/manifests/registry/machine.pp
@@ -105,6 +105,9 @@ class windows_template::registry::machine ()
     # operations to work, even though stricly speaking RunOnce is a function of explorer.exe and therefore should not
     # be supported on a Windows Core Installation.
     # Ticket https://tickets.puppetlabs.com/browse/IMAGES-577 has been raised for follow on investigation and work.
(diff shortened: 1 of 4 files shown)


jcoconnor:
This is ready for review, but I need to run this through build/test of ```win-2012r2-x86_64``` and ```win-2012r2-core-x86_64``` at a minimum.

jcoconnor:
So - initial testing on ```win-2012r2-core-x86_64``` shows an issue - the vmpooler post-clone configuration script doesn't run. 
If it is run manually, the post-clone configuration goes ok, including the new startup script.
```win-2012r2-x86_64``` operates correctly and without issue, i.e. post-clone works, bginfo triggered etc.
So wondering if the ```HKLM\\Software\\Microsoft\\Windows NT\\CurrentVersion\\Winlogon``` key is still required to kick off the initial run script for Core.
Checking logs and will update change if necessary.

jcoconnor:
```win-2012r2-core-x86_64``` is working this time with the "Run key hack" re-inserted.
So while this is a partial fix, I am happier using scheduled tasks to kick off the startup script rather than the Run Keys.

PR #342, opened 2018-06-22
(IMAGES-853) Disable First Logon Animation

jcoconnor:
Disabling this for Windows 10 as it delays the first login sequence.
All in the cause of reducing the machine prep time at first boot.

### DIFF
--- a/templates/win/common/scripts/common/windows-env.ps1
+++ b/templates/win/common/scripts/common/windows-env.ps1
@@ -209,19 +209,23 @@ Function Install_Win_Patch {
 # Helper Function to delete file, with try/catch to ignore errors.
 # This Function is used in both the clean host and clean-disk scripts.
 # Leaving Verbose options on in all cases so we can be certain files are being removed (IMAGES-684)
-Function ForceFullyDelete-Paths {
-  $filetodelete = $args[0]
-
+Function ForceFullyDelete-Path {
+  param(
+  [Parameter(Mandatory = $true, ValueFromPipeline = $true)]
+  [String]$Path,
+  [String]$LogFile
+  )
+  
   try {
-    if(Test-Path $filetodelete) {
-        Write-Output "Removing $filetodelete"
-        Takeown /d "$AnswerPromptYes" /R /f $filetodelete
-        Icacls $filetodelete /grant:r "*${WindowsAdminSID}:(OI)(CI)F" /t /c  2>&1
-        Remove-Item $filetodelete -Recurse -Force -Verbose
+    if(Test-Path $Path) {
+        Write-Output "Removing $Path" >> $LogFile 2>&1
+        Takeown /d "$AnswerPromptYes" /R /f $Path >> $LogFile 2>&1
+        Icacls $Path /grant:r "*${WindowsAdminSID}:(OI)(CI)F" /t /c >> $LogFile 2>&1
+        Remove-Item $Path -Recurse -Force >> $LogFile 2>&1
       }
     }
     catch {
-        Write-Output "Ignoring Error deleting: $filetodelete - Continue"
+        Write-Output "Ignoring Error deleting: $Path - Continue" >> $LogFile 2>&1
     }
 }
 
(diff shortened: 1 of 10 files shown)


jcoconnor:
The acceptance testing for the Vagrant images appears to be faulty at the moment, but initial indications are that the revised memory setting is working ok.

Continuing to work on test harness.

templates/win/common/scripts/common/windows-env.ps1
```diff
@@ -210,18 +210,21 @@ Function Install_Win_Patch {
...
+  )
+  
   try {
-    if(Test-Path $filetodelete) {
-        Write-Output "Removing $filetodelete"
-        Takeown /d "$AnswerPromptYes" /R /f $filetodelete
-        Icacls $filetodelete /grant:r "*${WindowsAdminSID}:(OI)(CI)F" /t /c  2>&1
-        Remove-Item $filetodelete -Recurse -Force -Verbose
+    if(Test-Path $Path) {
+        Write-Output "Removing $Path" >> $LogFile 2>&1
```
RandomNoun7: I think you might be able to shorten this call by moving the stream redirection from the end of the line over to the first redirect to the file. I think if it reads ```Write-Output "Removing $Path" *>> $LogFile``` you may get the intended effect without needing to  redirects. As long as this gets you the output you're looking for though feel free to ignore this commend as a nit.

templates/win/common/scripts/common/windows-env.ps1
```diff
@@ -210,18 +210,21 @@ Function Install_Win_Patch {
 # This Function is used in both the clean host and clean-disk scripts.
 # Leaving Verbose options on in all cases so we can be certain files are being removed (IMAGES-684)
 Function ForceFullyDelete-Paths {
```
RandomNoun7: Powershell functions should have names that are a verb followed by a singular noun. The assumption is that you could be handling either a single object or a collection of objects, but the noun should remain singular.

templates/win/common/scripts/common/windows-env.ps1
```diff
@@ -210,18 +210,21 @@ Function Install_Win_Patch {
...
-  $filetodelete = $args[0]
-
+  param(
+  [String]$Path,
```
RandomNoun7: If you add a little bit of syntax to this parameter declaration you can take the paths you want to delete from the pipeline. That will allow you to remove the ```| %{ ... "$_" ...}``` syntax below. Just add ```[Parameter(Mandatory = $true, ValueFromPipeline = $true)]``` above the parameter declaration, as per [this tutorial ](https://info.sapien.com/index.php/scripting/scripting-how-tos/take-values-from-the-pipeline-in-powershell) and it should work fine.

templates/win/common/scripts/common/windows-env.ps1
jcoconnor: Unfortunately the ```*>>``` syntax only appeared in PS 3.0, so need the ```2>&1``` bit (would have preferred to have the conciser format and did try it :) )

templates/win/common/scripts/provisioners/clean-disk-dism.ps1
```diff
@@ -89,7 +89,7 @@ Write-Output "Clearing Files"
     "$ENV:USERPROFILE\AppData\Local\Microsoft\Windows\WER\ReportQueue",
     "$ENV:ALLUSERSPROFILE\Microsoft\Windows\WER\ReportArchive",
     "$ENV:ALLUSERSPROFILE\Microsoft\Windows\WER\ReportQueue"
-) | % { ForceFullyDelete-Paths "$_" }
+) | % { ForceFullyDelete-Paths -Path "$_" -LogFile "$PackerLogs\clean-basefiles.log"}
```
RandomNoun7: If you take paths from the pipeline as noted above this would change to ``` | ForcefullyDelete-Path -LogFile "$PackerLogs\clean-basefiles.log"```

jcoconnor:
@RandomNoun7 Changes applied - is this ok ?

PR #348, opened 2018-07-26
(IMAGES-890) Fix PE Driver Paths for 2008/7 builds

jcoconnor:
Windows builds earlier than Win-8 (i.e. Win-2008, 2008R2 and 7) need to
use Vista Driver paths for the initial PE phase on vsphere-iso so that
the pvcsci drivers are loaded and the hard drive is visible.

Use similar method to what is used for Platform/firmware selection to
make this choice.

DriverPaths are only needed for the initial autounattend file and not
the post-clone one.


jcoconnor:
This appears to build the correct ```autounattend.xml``` files using the ```gen-xslt.sh```
Waiting to get a network issue resolved so I can verify its correctness for on the ```packer-vcenter-iso``` build

jcoconnor:
The generated ```autounattend.xml``` file is generated correctly sufficient to boot up ```windows-2008r2``` and ```windows-10``` - drivers are loading ok and remainder of install is starting.

So this is ready for review/merge.

PR #350, opened 2018-08-09
(WIP) (IMAGES-704) Merge Base and vagrant builds (vbox)

jcoconnor:
Merge the Base and Vagrant stage builds for Virtual Box into a single
build stage, with a copy of the original vagrant build called
"vagrant_base" to allow two stage build if ever needed.

This needs corresponding changes in the ci-jobs jjb configuration to do
the virtualbox build in a single step.

This removes the need for intermediate storage of the base builds which
is not necessary for the build of the builds as they are now stable.


jcoconnor:
This builds and passes ```facter``` acceptance testing on ```jenkins-imaging```
However - need to revise filename, description of stages.

PR #354, opened 2018-08-15
(IMAGES-818) Update SLES15 template to final GM release

ScottGarman:
Includes additional changes:
* SLES 15 requires signed repos by default, and the localmirror we use
generated from the ISOs is not signed, so the gpgcheck needed to be
disabled
* De-register the system using the SUSEConnect command to clean up the
upstream repo configs instead of deleting them manually
* Disable the use of client-id for DHCP4 requests, as template clones
resulted in client-id collisions


manifests/modules/packer/manifests/vsphere/repos.pp
```diff
@@ -195,20 +195,34 @@
...
         enabled     => 1,
         autorefresh => 1,
         baseurl     => "${base_url}",
-        gpgcheck    => "1",
+        gpgcheck    => "${gpg_check}",
         gpgkey      => "file:///etc/pki/rpm-gpg/${gpgkey}",
         type        => 'rpm-md'
       }
+
+      if $::operatingsystemrelease == "15.0" {
```
ekinanp: Nice, much better solution than what I had.

PR #362, opened 2018-09-04
(IMAGES-915) Add Mock-Platform build for JJB testing

jcoconnor:
This is a partial "quick" windows platform build with enough
components, i.e. Cygwin and ssh settings but nothing else to
provide a fast build template for JJB testing.

### DIFF
--- a/templates/win/common/scripts/bootstrap/start-pswindowsupdate.ps1
+++ b/templates/win/common/scripts/bootstrap/start-pswindowsupdate.ps1
@@ -168,6 +168,10 @@ Import-Module "$PackerStaging\PSWindowsUpdate\PSWindowsUpdate.psd1"
 # Windows-10 in particular seems to be affected by intermittency here - so try and improve reliability
 $Attempt = 1
 do {
+  if (Test-Path "$PackerLogs\Mock.Platform" ) {
+    Write-Output "Test Platform Build - exiting"
+    break
+  }
   Write-Output "Windows Update Pass $Attempt"
   try {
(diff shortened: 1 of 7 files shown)


jcoconnor:
This builds the Mock 2019 platform image in about 10 minutes.
Have also tested a standard build and it works fine too.

So ready to Merge methinks.

templates/win/common/scripts/bootstrap/start-pswindowsupdate.ps1
```diff
@@ -168,6 +168,10 @@ Import-Module "$PackerStaging\PSWindowsUpdate\PSWindowsUpdate.psd1"
 # Windows-10 in particular seems to be affected by intermittency here - so try and improve reliability
 $Attempt = 1
 do {
+  if (Test-Path "$PackerLogs\Mock.Platform" ) {
```
ekinanp: Could we pass in an boolean environment variable instead? E.g. `PACKER_BUILD_MOCK_PLATFORM`? Would avoid the extra script for creating the file.

jcoconnor: Environment variables aren't preserved across restarts, so have kind of settled on "marker" touch-type files.

This is kind of a quick way at disabling the longer parts of the build (mainly windows update and dism cleanup), while leaving enough of the build to allow cygwin/ssh to spin up so we can have reasonably quick pipeline for testing.

ekinanp: Cool, no problem. Thanks for clarifying the reason.

PR #363, opened 2018-09-07
(IMAGES-917) Seperate the bootstrap stage from win update

jcoconnor:
If a failure happens in the bootstrap phase, it won't get notified until
after a very long timeout period (several hours). So minimise the bootstrap
phase to just installing the hypervisor tools/drivers and any other minimal
network operations so that WinRM can be enabled and control passed back to
packer.

The Windows update phase is initiated later in a seperate phase using the
technique successfully used with the Appveyor build, where WinRm is disabled
just before a packer-reboot provider to allow a further reboot cycle. This
cycle is ended by re-enabling WinRM - thus allowing Windows Update to use as
many reboots as needed.

Also - have changed the startup/reboot sequence to use a Scheduled Task on
Startup to continue the sequence across multiple reboots. This has the
added advantage of being able to run the script elevated, thus getting
round the restriction that WSUS cannot be used across WinRM sessions.

It has also allowed the removal of the $PROFILE Core edition workaround.

This gives a more immediate experience to the jenkins log that the boot sequence
has completed at least.

### DIFF
--- a/templates/win/common/scripts/bootstrap/bootstrap-packerbuild.ps1
+++ b/templates/win/common/scripts/bootstrap/bootstrap-packerbuild.ps1
@@ -1,13 +1,21 @@
+# Script to perform the minimal actions needed to bring the host online with winrm
+# i.e, install the hypervisor tools, network adaptor mods and necessary reg fixes.
+# The windows update is deferred until a later stage.
+# This means that more immediate feedback is given to packer that the OS load itself
+# is sound.
+
 param (
-    [string]$HyperVisor = "vmware"
+    [string]$HyperVisor = "vmware",
+    [string]$AdminUser = "Administrator",
+    [string]$AdminPassword = "PackerAdmin"
 )
 $ErrorActionPreference = 'Stop'
 
 . A:\windows-env.ps1
 $PackageDir = 'A:\'
 
-$rundate = date
-write-output "Script: start-pswindowsupdate.ps1 Starting at: $rundate"
+$rundate = Get-Date
+Write-Output "Script: bootstrap-packerbuild.ps1 Starting at: $rundate"
 
 # Create Packer Log Directories if they don't exist already.
 Create-PackerStagingDirectories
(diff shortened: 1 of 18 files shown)


templates/win/common/scripts/bootstrap/bootstrap-packerbuild.ps1
```diff
@@ -240,7 +171,9 @@ winrm set winrm/config/winrs '@{MaxMemoryPerShellMB="2048"}'
 # Set service to start automatically (not delayed)
 Set-Service "WinRM" -StartupType Automatic
 
-Write-Output "WinRM setup complete"
 
-# Clear reboot files as control is now transferred to Packer to complete configuration.
-Clear-RebootFiles
+# Bootstrap Cycle complete - delete this task.
+Write "Deleting Bootstrap Scheduled Task"
```
RandomNoun7: Is this intented to be Write-Output? Write-Host?

templates/win/common/scripts/bootstrap/packer-windows-update.ps1
```diff
@@ -0,0 +1,89 @@
...
+}
+
+# Run Windows Update - this will repeat as often as needed through the Invoke-Reboot cycle.
+# When no more reboots are needed, the script falls through to the end.
+Write-Output "Searching for Windows Updates"
+if ($WindowsVersion -like $WindowsServer2016) {
+  Write-Output "Disabling some more Windows Update (10) parameters"
+  Write-Output "Disable seeding of updates to other computers via Group Policies"
+  force-mkdir "HKLM:\SOFTWARE\Policies\Microsoft\Windows\DeliveryOptimization"
+  Set-ItemProperty "HKLM:\SOFTWARE\Policies\Microsoft\Windows\DeliveryOptimization" "DODownloadMode" 0
```
RandomNoun7: Consider using named instead of positional parameters.

templates/win/common/scripts/bootstrap/packer-windows-update.ps1
```diff
@@ -0,0 +1,89 @@
...
+} while ($Attempt -le 2)
+
+# Run the Application Package Cleaner
+if (Test-Path "$PackerLogs\AppsPackageRemove.Required") {
+  Write-Output "Running Apps Package Cleaner post windows update"
+  Remove-AppsPackages -AppPackageCheckpoint AppsPackageRemove.Pass1
+}
+
+# Windows Update Cycle complete - delete this task.
+Write "Deleting Bootstrap Scheduled Task"
```
RandomNoun7: I see now this is an alias for write-output, but aliases should probably be avoided in code committed to repo's.

templates/win/common/scripts/common/windows-env.ps1
```diff
@@ -553,60 +538,38 @@ Function Test-PendingReboot {
...
   # Drop out and don't reboot.
-  Write-Host -ForegroundColor Green "Reboot is not needed"
+  Write-Host "Reboot is not needed"
   return $false
 }
 
 # Helper Function to perform a reboot and continue the windows update process.
 Function Invoke-Reboot {
+    param(
+      $NoReboot=$false )
```
RandomNoun7: Remove?

templates/win/common/scripts/provisioners/initiate-windows-update.ps1
```diff
@@ -0,0 +1,44 @@
+
+# Initiate the Windows Update operation.
+
+$ErrorActionPreference = 'Stop'
+
+. C:\Packer\Scripts\windows-env.ps1
+
+Write-Output "Setting up Windows Update"
+
+# Important Pre-requisite right across the packer  including Windows Update.
```
RandomNoun7: Double space after packer.

jcoconnor: ok

RandomNoun7:
Comments Complete.

templates/win/common/scripts/common/windows-env.ps1
jcoconnor: Yip

jcoconnor:
This appears to build a good sample of the main ```vCenter``` without issue.
The only persistent issue was the ```win-2012-x86_64``` build with reboots interrupting the win-packages provider. However, the latest build for this ran clean.

Propose that this is merged so that we can do a full PROD build tonight once the Sept Patch Tuesdays patches are out.

templates/win/common/scripts/bootstrap/packer-windows-update.ps1
jcoconnor: Good suggestion - may defer to the next PR.

PR #364, opened 2018-09-07
PUP-9084 Added vars.json for Ubuntu 18.10 x86_64

sebastian-miclea:
Added config files for Ubuntu 18.19 x86_64 platform.
This turned out as being useful as it can help test newer versions of Ubuntu before the official release.

### DIFF
--- a/templates/ubuntu/18.10/x86_64/vars.json
+++ b/templates/ubuntu/18.10/x86_64/vars.json
@@ -1,0 +1,12 @@
+{
+    "template_name"                         : "ubuntu-18.10-x86_64",
+    "template_os"                           : "ubuntu-64",
+    "beakerhost"                            : "ubuntu1810-64",
+    "version"                               : "0.0.1",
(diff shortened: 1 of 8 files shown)


templates/ubuntu/18.10/x86_64/vars.json
```diff
@@ -0,0 +1,12 @@
+{
+    "template_name"                         : "ubuntu-18.10-x86_64",
+    "template_os"                           : "ubuntu-64",
+    "beakerhost"                            : "ubuntu1804-64",
+    "version"                               : "0.0.2",
```
caseywilliams: This all looks good, but I think the version should be 0.0.1

sebastian-miclea: You're right, nice catch!
Updated, thanks

mihaibuzgau:
Image is published on prod vmpooler

PR #367, opened 2018-09-18
(IMAGES-859) Update win-10-ent to 1803

jcoconnor:
Build Win-10 1803 images as the main win-10-ent images as these
the beaker upgrade issue in the above ticket is no longer evident.

Remove the 1709 and 1803 builds.


jcoconnor:
This has been built successfully and deployed on vmpooler - however, holding off merging for 24 hours until certain that there are no adverse effects on pipelines.

PR #368, opened 2018-09-20
(IMAGES-918) Localise Admin User/Password

jcoconnor:
Issue with French/German localised Admin names for the scheduled Task work.
Fix so that we can pass the correct Username/Password into the scripts.

### DIFF
--- a/templates/win/common/files/Autounattend.xslt
+++ b/templates/win/common/files/Autounattend.xslt
@@ -1,10 +1,9 @@
 <xsl:stylesheet version="1.0"
-            xmlns="http://www.w3.org/1999/xhtml"
             xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
             xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State"
             xmlns:u="urn:schemas-microsoft-com:unattend">
 
-  <xsl:output method="xml" indent="yes" encoding="utf-8" />
+  <xsl:output method="xml" indent="yes" encoding="utf-8" omit-xml-declaration="yes" doctype-system="about:legacy-compat"/>
 
   <xsl:strip-space elements="*"/>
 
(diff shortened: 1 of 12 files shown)


templates/win/common/files/Autounattend.xslt
```diff
@@ -84,11 +84,57 @@
     </xsl:copy>
   </xsl:template>
 
-  <!-- Rule to select appropriate logon sequence -->
+  <!-- Rule to select appropriate logon sequence and perform Administrator/Password substitutions. -->
   <xsl:template match='u:unattend/u:settings/u:component[@name="Microsoft-Windows-Shell-Setup"]/u:PackerLogonSequence'>
        <xsl:if test="@ImageProvisioner=$ImageProvisioner">
           <!-- Strip out PackerLogonSequence once selected to present valid Unattend.xml -->
-          <xsl:copy-of select="node()" />
+          <xsl:for-each select="node()">
```
Iristyle: Can't you use a more direct `match` to get the right nodes picked out rather than looping?

jcoconnor: Tx much for the pointers will absorb them tomorrow.
The good news is that with the minor seq fix below, this appears to come up ok for the EN-us case and also fixes the issue I noticed with Win-2019 ([IMAGES-928](https://tickets.puppetlabs.com/browse/IMAGES-928))

jcoconnor:
@Iristyle I've updated the template handling as per your suggestions.
I have one more fix that possibly could be used to remove the <xsl:choose> from the CommandLine handling rule.

jcoconnor:
Work for this is essentially complete.
Pushing through a full set of builds over the weekend to validate the work.

jcoconnor:
Hmm - spoke too soon - authentication issue with French only - will look at on Monday.

jcoconnor:
OK - Issue with French resolved - appears to be building ok now.

jcoconnor:
All images built and ran acceptance tests overnight, so French issue above is sorted.

jcoconnor:
Still need a small fix to the French Post-Clone startup script. Doing......

jcoconnor:
Ok - this is building and running the acceptance suite on the English and French platforms.
So ready to merge pending review.

PR #369, opened 2018-10-01
[IMAGES-924] add solaris 11.4

loopinu:
This PR adds support for Solaris 11.4.

### DIFF
--- a/manifests/modules/packer/manifests/vmtools.pp
+++ b/manifests/modules/packer/manifests/vmtools.pp
@@ -25,23 +25,50 @@ class packer::vmtools inherits packer::vmtools::params {
       ensure => directory,
     }
 
-    mount { '/tmp/vmtools':
-      ensure  => mounted,
-      device  => "${root_home}/${tools_iso}",
-      fstype  => 'iso9660',
-      options => 'ro,loop',
-      require => File[ '/tmp/vmtools' ],
+    if $::osfamily == 'Solaris' {
+      mount { '/tmp/vmtools':
+        ensure      => mounted,
+        device      => "${root_home}/${tools_iso}",
+        fstype      => 'hsfs',
+        remounts    => false,
+        blockdevice => '/dev/rdsk/c0d0s0',
+        atboot      => true,
+        options     => 'ro,loop',
+        require     => File[ '/tmp/vmtools' ],
+      }
+    } else {
+      mount { '/tmp/vmtools':
+        ensure  => mounted,
+        device  => "${root_home}/${tools_iso}",
+        fstype  => 'iso9660',
+        options => 'ro,loop',
+        require => File[ '/tmp/vmtools' ],
+      }
     }
 
     exec { 'install vmtools':
       command => $install_cmd,
-      path    => [ '/bin', '/usr/bin', ],
+      path    => [ '/bin', '/usr/bin', '/sbin', '/usr/sbin' ],
+      cwd     => '/tmp/',
       require => Mount[ '/tmp/vmtools' ],
+    }
+
+    if $::osfamily == 'Solaris' {
+      # this is required because the vmware-tools installation fails with 0 exit code
+      # the next step tries to validate that the service is running.
+      service { 'vmware-tools':
+        ensure   => running,
+        require  => Exec[ 'install vmtools' ],
+        start    => '/etc/init.d/vmware-tools start && /etc/init.d/vmware-tools status',
+        stop     => '/etc/init.d/vmware-tools stop',
+        status   => '/etc/init.d/vmware-tools status',
+        provider => base
+      }
     }
 
     exec { 'remove /tmp/vmtools':
       command => 'umount /tmp/vmtools ; rmdir /tmp/vmtools',
-      path    => [ '/bin', '/usr/bin', ],
+      path    => [ '/sbin', '/usr/sbin', '/bin', '/usr/bin' ],
       onlyif  => 'test -d /tmp/vmtools',
       require => Exec[ 'install vmtools' ],
     }
@@ -51,8 +78,14 @@ class packer::vmtools inherits packer::vmtools::params {
       require => Exec[ 'remove /tmp/vmtools' ],
     }
 
-    file_line { "remove /etc/fstab /tmp/vmtools":
-      path    => '/etc/fstab',
+    if $::osfamily == 'Solaris' {
+      $fstab_path = '/etc/vfstab'
+    } else {
+      $fstab_path = '/etc/fstab'
+    }
+
+    file_line { 'remove fstab /tmp/vmtools':
+      path    => $fstab_path,
       line    => '#/tmp/vmtools removed',
       match   => '/tmp/vmtools',
       require => Exec[ 'remove /tmp/vmtools' ],
--- a/manifests/modules/packer/manifests/vsphere.pp
+++ b/manifests/modules/packer/manifests/vsphere.pp
@@ -57,9 +57,9 @@ class packer::vsphere inherits packer::vsphere::params {
   }
 
   file { $bootstrap_file:
-    owner  => 'root',
-    group  => 'root',
-    mode   => '0755',
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0755',
     content => template("packer/vsphere/${bootstrap_file_source}"),
   }
 
@@ -71,9 +71,9 @@ class packer::vsphere inherits packer::vsphere::params {
   }
 
   file { '/root/.ssh':
-    owner   => 'root',
-    group   => 'root',
-    mode    => '0755',
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0755',
     ensure => directory,
   }
 
@@ -83,6 +83,20 @@ class packer::vsphere inherits packer::vsphere::params {
     mode    => '0755',
     source  => 'puppet:///modules/packer/vsphere/authorized_keys',
     require => File[ '/root/.ssh' ]
+  }
+
+  #TODO check if this works with existing template for solaris 11
+  if $::operatingsystem == 'Solaris' {
+    if $::operatingsystemrelease in ['11.4'] {
+      file { "/etc/rc2.d/S99${startup_file_source}":
+          ensure => 'link',
+          target => $startup_file,
+        }
+      file { "/etc/rc0.d/K99${startup_file_source}":
+          ensure => 'link',
+          target => $startup_file,
+        }
+    }
   }
 
 }
--- a/manifests/modules/packer/manifests/vsphere/networking.pp
+++ b/manifests/modules/packer/manifests/vsphere/networking.pp
@@ -47,6 +47,6 @@ class packer::vsphere::networking inherits packer::networking::params {
           source => 'puppet:///modules/packer/vsphere/dhcp4.xml',
         }
       }
-    }
+    }    
   }
 }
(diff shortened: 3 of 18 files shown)


manifests/modules/packer/manifests/vmtools.pp
```diff
@@ -51,8 +78,14 @@
       require => Exec[ 'remove /tmp/vmtools' ],
     }
 
-    file_line { "remove /etc/fstab /tmp/vmtools":
-      path    => '/etc/fstab',
+    if $::osfamily == 'Solaris' {
+      $fsdab_path = '/etc/vfstab'
+    } else {
+      $fsdab_path = '/etc/fstab'
```
ScottGarman: Is $fstab_path an already used variable name? I'd prefer to use that if we can.

ScottGarman:
Minor nit: the commit header should be in parenthesis, not square brackets. (IMAGES-924)

loopinu:
Created new directory for Solaris 11.4 
Added new ai.xml definitions for solaris 11.4 auto installer
Added profile.xml configuration to set up root account and password, network configuration for dhcp and dns etc.

Added i386/vars.json with the required variables. Note( we don't use puppet_aio variable and instead we use pc_repo) for bootstrap-ai.sh

Added new condition's in bootstrap-aio.sh and cleanup-aio.sh to execute the install/uninstall of puppet for Solaris

Added new resource in vmtools/params.pp manifest for Solaris also modified vmware resource with a condition in case osfamily is solaris to execute install_cmd with 'tar zxf /tmp/vmtools/vmware-solaris-*.tar.gz instead of 'tar zxf /tmp/vmtools/VMwareTools-*.tar.gz.

Also in vmtools/params.pp we added service resource if osfamily is Solaris to check if vmware tools is installed and running this is required because the vmware-tools installation failes with 0 exit code.

Added new resorce in updates.pp manifest for Solaris to include openssh and kernel packages and update them.

Added new resorce in networking/params.pp to include support for Solaris.

Modified vmware.base.json boot_command to support .HTTIP and .HTTPORT in vars.json that can defined at the OS/OSDistribution/Architecture this was needed because we couldn't use the curl command to download the ai.xml and profile.xml from http_directory because packer does not support string interpolation in vars.json. The mecanism is used in combination with export PACKURL=%s in witch %s is parsed as the httpip:/httpport from packer and after that we use it for curl multiple times in the boot_command. NOTE this will not affect other boot_commands only if in the boot_command will use (%s).

mihaibuzgau:
@ScottGarman do you want us to squash this commits ?

ScottGarman:
Yes. Please squash these into one commit. Thanks!

templates/common/vmware.base.json
```diff
@@ -53,12 +52,12 @@
...
       "guest_os_type"                              : "{{user `template_os`}}",
       "disk_size"                                  : "{{user `disk_size`}}",
-      "tools_upload_flavor"                        : "linux",
+      "tools_upload_flavor"                        : "{{user `tools_upload_flavor`}}",      
       "floppy_dirs"                                : "{{user `floppy_dirs`}}",
       "floppy_files"                               : "{{user `floppy_files`}}",
       "boot_wait"                                  : "{{user `vmware_base_boot_wait`}}",
-      "boot_command"                               : "{{user `boot_command`}}",
-
+      "boot_command"                               : "{{$flag:= user `inject_http_seed_in_boot_command`}}{{$bootCmd:= user `boot_command`}}{{if eq $flag `true`}}{{$seed:= printf \"%s:%d\" .HTTPIP .HTTPPort}}{{$seed | printf $bootCmd}}{{else}}{{$bootCmd}}{{end}}",
```
ekinanp: This is interesting. Will this break any existing templates? If not, it would be great to refactor some of our existing templates to use this format (e.g. our Debian ones) [that work can be done separately from this PR]. It may be worthwhile to run some of our existing imaging pipelines just to see if we can build a base template. Perhaps one of each OS that uses the top-level template (e.g. CentOS, Fedora, RHEL).

ekinanp: If things don't break, then I think it would be better to just remove the flag and have it pass along the HTTPIP HTTPPort as environment variables. Can you do something like
```
export PACKER_HTTPIP = .HTTPIP
export PACKER_HTTP_PORT = .HTTPPort
```

in the boot command?

scripts/bootstrap-aio.sh
```diff
@@ -45,17 +49,22 @@ else
 fi
 
 # Show Puppet version
-printf 'Puppet ' ; /opt/puppetlabs/puppet/bin/puppet --version
+if which puppet ; then
+  PUPPET_CMD=puppet
+else
+  PUPPET_CMD=/opt/puppetlabs/puppet/bin/puppet
```
ekinanp: What is the reason for this variable? Won't `/opt/puppetlabs/puppet/bin/puppet` always work?

scripts/bootstrap-aio.sh
```diff
@@ -22,6 +22,10 @@ if [ -n "${PC_REPO}" ]; then
     dpkg -i $tmp_dir/pc-repo.deb
     apt-get update
     apt-get install -y puppet-agent
+  elif type pkg >/dev/null ; then
+      pkg install puppet
+      svccfg -s puppet:agent setprop config/server=master.oracle.com
```
ekinanp: Why `svccfg` here?

manifests/modules/packer/manifests/vsphere/repos.pp
```diff
@@ -224,10 +224,16 @@
         }
       }
     }
+    solaris: {
+      if $::operatingsystemrelease in ['11.4'] {
```
ekinanp: Can this code be removed?

manifests/modules/packer/manifests/vsphere/params.pp
```diff
@@ -95,6 +95,16 @@
       $gpgkey                = "RPM-GPG-KEY-${::operatingsystemmajrelease}-${loweros}"
     }
 
+    'Solaris': {
+      if $::operatingsystemrelease in ['11.4'] {
```
ekinanp: May be worthwhile to have a `TODO` to see if this would work with the main Solaris 11 template?

manifests/modules/packer/manifests/vsphere/networking.pp
```diff
@@ -48,5 +48,9 @@
         }
       }
     }
+
+    solaris: {
```
ekinanp: Let's remove this code if it is not needed.

manifests/modules/packer/manifests/vsphere.pp
```diff
@@ -85,4 +85,17 @@
     require => File[ '/root/.ssh' ]
   }
 
+  if $::operatingsystem == 'Solaris' {
```
ekinanp: Maybe a `TODO` here to see if it works with the existing Sol 11 template?

manifests/modules/packer/manifests/vmtools.pp
```diff
@@ -25,23 +25,50 @@
...
     exec { 'install vmtools':
       command => $install_cmd,
-      path    => [ '/bin', '/usr/bin', ],
+      path    => [ '/bin', '/usr/bin', '/sbin', '/usr/sbin' ],
+      cwd     => '/tmp/',
       require => Mount[ '/tmp/vmtools' ],
     }
 
+    if $::osfamily == 'Solaris' {
+      #this is required because the vmware-tools installation failes with 0 exit code
```
ekinanp: Typos in this comment + space after the `#` (e.g. top-line should be `# this is required ...`)

templates/common/vmware.base.json
ekinanp: Perhaps `export PACKER_SERVER_URL` would be better.

loopinu: The flag is required because %s is interpreted by the go sprintf(printf in the template) function and not by the templating engine (vars.json does not pass trough any templating engine) and if we do not have a %s (or any other replacement symbol) in the string the sprintf function will complain that it has a missing param and will output unintentional text in the boot command(eg. MISSING AGRUMENT x). What the flag does is ensures this never happens unintentionally. 

The current implementation injects .HTTPIP:.HTTPPort in the vars file as %s which you can use to export any environment var (check vars.json in solaris 11.4)

scripts/bootstrap-aio.sh
loopinu: For Solaris 11.4 it points to /usr/sbin/puppet that's why we added this code

manifests/modules/packer/manifests/vsphere.pp
loopinu: Done.

manifests/modules/packer/manifests/vsphere/params.pp
loopinu: Done.

templates/common/vmware.base.json
ekinanp: That makes sense -- thanks for the explanation. Nice, clever use of Go's templating syntax too! I've found that this also works:
```
{{ $PACKER_SERVER_URL := printf \"%s:%d\" .HTTPIP .HTTPPort }} {{ printf (user `boot_command`) $PACKER_SERVER_URL }}
```

Writing it this way will work even when the boot command does not have a `%s` so it looks like we do not need the flag.

ekinanp: Unfortunately it looks like introducing the extra Go-lang template interpolation code will break some of our existing images (e.g. Fedora 28). I'm not sure why this happens, but it happens. For example,
```
{{ $PACKER_SERVER_URL := printf \"http://%s:%d\" .HTTPIP .HTTPPort }} {{ user `boot_command` }}
```

broke Fedora 28 (notice there's no `printf` there before the boot command). I hate to admit it but if you can verify that your changes here will also break Fedora 28, then we will need to create a separate `vmware.base` template for Solaris 11.4. It would be good to file a separate ticket, however, to investigate and see whether we can get the templates working via. a single boot command now that we've seen that it can be done for some platforms (Solaris 11.4, Debian).

loopinu: I tested Fedora 28, Debian 9.0, RHEL 7.2 and Ubuntu 18.04 it works as expected with the current implementation as it is. I suggest we leave it like this and when we want to use HTTPIP & HTTPPort we can introduce "inject_http_seed_in_boot_command"      : "true", in the specific os vars.json and use the export PACKURL=%s in the os boot command.

ekinanp: Nevermind, it does work. My code had spaces between the template interpolations, so that space was interpreted as being the first character of the boot command. Otherwise, I've also verified it works. The only other suggestion I have is let us change the boot command in the top-level template to this:
```
{{ $packer_server_url := printf \"http://%s:%d\" .HTTPIP .HTTPPort }}{{ printf (user `boot_command`) $packer_server_url }}
```

This way, a boot_command can retrieve the server_url via. `%s`, but packer will still type in the right boot command if the boot command does not have a `%s`. This should eliminate the need for the flag.

ekinanp: We should also file a ticket to remove the use of `floppy` from our templates, as the only reason that was introduced was because I originally thought that you could not include the `HTTP` stuff in a boot command when the boot command is in a user variable (nested interpolation) which is still correct, however it is nice to see there's a clever way to workaround that. Once again, nice work on the boot command!

scripts/bootstrap-aio.sh
loopinu: This configures puppet agent and it's needed at nocm step, with out this configuration nocm step will exit with non-zero exit status: 1.

templates/common/vmware.base.json
loopinu: I tried to Fedora 28 with {{ $packer_server_url := printf \"http://%s:%d\" .HTTPIP .HTTPPort }}{{ printf (user `boot_command`) $packer_server_url }} the problem is still there MISSING AGRUMENT x you can check this by enabling packer_log=1

PR #370, opened 2018-10-01
(IMAGES-909) Set firmware for vCenter builds to efi

jcoconnor:
Point the builder to use the new "puppet" build of the
packer-build-vsphere-iso plugin (which is deliberately
named as packer-build-vsphere-plugin-iso to disinguish that
we are using a custom puppet built one until such time as
jetbrains release a revised plugin.

Having the new plugin in place means that we can now direct
the firmware for vcenter builds to efi (except for win-2008
of course). The firmware setting is also removed from the
vmx parameters, where it appeared to cause problems with
ovftool and esxi 6.7

### DIFF
--- a/templates/win/common/vars.json
+++ b/templates/win/common/vars.json
@@ -1,5 +1,5 @@
 {
-  "version"             : "20180920_DEV",
+  "version"             : "20181003_DEV",
 
   "headless"            : "true",
 
--- a/templates/win/common/vmware.vcenter.cygwin.json
+++ b/templates/win/common/vmware.vcenter.cygwin.json
@@ -40,7 +40,7 @@
 
   "builders": [
     {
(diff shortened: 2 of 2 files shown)


jcoconnor:
Running through a full set of builds.
Initial results look positive - all new images are booting up as EFI.

This is ready for review/merge.

jcoconnor:
This ticket is based on the changes in: https://github.com/jetbrains-infra/packer-builder-vsphere/pull/156

A local build has been made of the plugin which is installed on all the build machines at: {{~jenkins/.packer.d/plugins}} as well as in the common directory {{/src/packer/tools/packer-plugins}}

jcoconnor:
All builds passed ok on a full run last night.

PR #373, opened 2018-10-12
(maint) Add Solaris to bypass default case in repos.pp

loopinu:
This is needed to pass nocm and to not enter the default statement.

### DIFF
--- a/manifests/modules/packer/manifests/vsphere/repos.pp
+++ b/manifests/modules/packer/manifests/vsphere/repos.pp
@@ -224,7 +224,11 @@ class packer::vsphere::repos inherits packer::vsphere::params {
         }
       }
     }
-    
+
+    solaris: {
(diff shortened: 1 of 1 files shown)


manifests/modules/packer/manifests/vsphere/repos.pp
```diff
@@ -224,7 +224,10 @@
         }
       }
     }
-    
+
+    solaris: {
```
ekinanp: I'd probably word the comment here to indicate that we need this statement so that the manifest won't fail for Solaris, otherwise it seems like we already have our Solaris repos set up?

loopinu: I agree with you, just added second comment to indicate the purpose of it.

ScottGarman:
:+1:

PR #377, opened 2018-10-24
(maint) Update SLES 15 agent to use official released version

ScottGarman:
During image development for SLES 15, I had to create a temporary agent
package that included important fixes in facter to detect SLES 15
correctly. Now that we've released our first official puppet5 package
for SLES 15, this updates the image build to use that package instead.


ScottGarman:
Note: this won't cause any difference in the VM templates that are generated, so there was no need to bump the template version.

PR #381, opened 2018-10-26
Ubuntu 14.04 should use vmware-tools instande of openvmtools

mihaibuzgau:
It seems that open-vm-tools causes Ubuntu 14.04 to error out in vmpooler. This PR removes the dependency on open-vm-tools and enables VMWare tools


ekinanp:
Has this been tested and deployed via. the imaging pipelines?

PR #383, opened 2018-10-30
(IMAGES-691) Fix OSX 10.12 templates

loopinu:
Modified osx 10.12 to use the vmware.base and nocm like the rest of the linux distros.

Modified manifests to support Dawin os type.

Modified boostrap_aio script to install puppet agent & xcode-commandline-tools for macos 

Added new vmx variables as user variables and added default values for them to be able to use common vmware.base file for installation of osx:


ekinanp:
Nice!

manifests/modules/packer/manifests/vsphere.pp
```diff
@@ -51,40 +62,77 @@
       }
     }
   }
+  # Added below block because in Darwin os the paths are different from the other os
+  # Darwin os has ruby installed by default
+    if $::osfamily == 'Darwin'{
```
ekinanp: Could clean this up as
```
if osfamily == Darwin {
    $startup_file_plist
} else {
   $ruby_package
}

bootstrap_file
startup_file
.ssh directory
ssh authorized_keys
```

scripts/bootstrap-aio.sh
```diff
@@ -60,8 +77,13 @@ printf 'Puppet ' ; $PUPPET_CMD --version
 # Installed required modules
 for i in $@
 do
+  if [[ ${PUPPET_AIO} == *".dmg"* ]]; then
+    chown -R osx /tmp/packer-puppet-masterless/manifests/modules
+    $PUPPET_CMD module install $i --modulepath=/tmp/packer-puppet-masterless/manifests/modules
```
ekinanp: Why is this command repeated twice for OSX? And couldn't we move the chown to outside the loop?

templates/common/vmware.base.json
```diff
@@ -58,13 +59,22 @@
...
         "memsize"                                  : "{{user `vmware_base_vmx_data_memsize`}}",
         "cpuid.coresPerSocket"                     : "{{user `vmware_base_vmx_data_cpuid_coresPerSocket`}}",
         "numvcpus"                                 : "{{user `vmware_base_vmx_data_numvcpus`}}",
-
+        "keyboardAndMouseProfile"                  : "{{user `vmware_base_vmx_keyboard_and_mouse_profile`}}",
         "ethernet0.virtualDev"                     : "{{user `vmware_base_vmx_data_ethernet0_virtualDev`}}",
-        "ethernet0.pciSlotNumber"                  : "{{user `vmware_base_vmx_data_ethernet0_pciSlotNumber`}}"
+        "ethernet0.pciSlotNumber"                  : "{{user `vmware_base_vmx_data_ethernet0_pciSlotNumber`}}",
+        "ehci.present"                             : "{{user `vmware_base_vmx_ehci_present`}}",
+        "firmware"                                 : "{{user `vmware_base_vm_firmware`}}",
```
ekinanp: Don't we want default values for these variables? I don't see them here. [Packer's source](https://github.com/hashicorp/packer) should have them available as that's where I got the original ones in the file.

manifests/modules/packer/manifests/vsphere.pp
loopinu: done.

scripts/bootstrap-aio.sh
loopinu: Good catch, removed the extra lines because we don't need it anymore.

templates/common/vmware.base.json
loopinu: I added them here to be able to remove macos base.json and to be able to set the variables in vars.json. I generated two images (ubuntu & fedora) with success using this modification but there is a catch in the vmx file generated the variables are set to empty string and before this the vmx file did't generate those variables at all. I need to investigate this a little bit more.

loopinu: @ekinanp I did some research on what values are written in the vmx file. If we don't define the variables in the vars.json and there is no default value in base.json the property from vmx file will be set as an empty string. ex: firmware="". I added some default values that the vmx file was generating before I added those variables. ex:
 "vmware_base_vmx_ehci_present"                 : "FALSE",
  "vmware_base_vm_usb_present"                   : "FALSE" ,
  "vmware_base_vm_scsi0_present"                 : "TRUE",
The rest of the variables that where added will be set as empty string in the vmx file but I don't think this will have an impact on the rest of the templates because before this those properties where not generated at all.
Also I did some local testing for ubuntu, fedora and redhat and they work as expected.

loopinu:
@ekinanp I tried to make an experimental pipeline to run the steps of building the image on jenkins but with no success because the agent that tries to build has centos with vmware workstation but for osx we need to run it on mac mini or vsphere. The plan is to make the pipeline work on vSphere but until then can we merge this PR?

ekinanp:
@loopinu Yes, I'm fine merging this. Can you change the commit messages to be more consistent with what's in the PR title? E.g.
```
<ticket_no> <summary>

<description>
```

PR #384, opened 2018-11-06
(IMAGES-946) Added template for Fedora 29 x86-64

sebastian-miclea:


templates/fedora/29/x86_64/vars.json
```diff
@@ -0,0 +1,14 @@
...
+    "beakerhost"                            : "fedora29-64",
+    "version"                               : "0.0.1",
+    "iso_url"                               : "https://artifactory.delivery.puppetlabs.net/artifactory/generic/iso/Fedora-Server-dvd-x86_64-29-1.2.iso",
+    "iso_checksum"                          : "129d131a55e5bd518f593f0eacdce095f7c795fe7ccbef1f3f6aeb2ff9f99f35",
+    "iso_checksum_type"                     : "sha256",
+    "vmware_vsphere_nocm_vmx_data_memsize"  : "4096",
+    "vmware_vsphere_nocm_vmx_data_numvcpus" : "2",
+    "floppy_files"                          : "../files/ks.cfg",
+    "puppet_aio"                            : "http://yum.puppetlabs.com/fedora/f27/PC1/x86_64/puppet-agent-1.10.12-1.fedoraf27.x86_64.rpm",
+    "pc_repo"                               : ""
```
ekinanp: Why is this needed?

manifests/modules/packer/manifests/networking/params.pp
```diff
@@ -34,6 +34,12 @@
           $udev_rule_gen = undef
         }
 
+        '29': {
+          case $::provisioner {
+            'libvirt':    { $interface_script = '/etc/sysconfig/network-scripts/ifcfg-ens33' }
```
ScottGarman: I think this will need to be updated with a case for vmware (virtualbox would be nice too, but isn't mandatory or officially supported by us)

templates/fedora/29/x86_64/vars.json
sebastian-miclea: That was left behind in vmware.vsphere.nocm.json as null which made it required.
I removed it from both places.

Thanks!

manifests/modules/packer/manifests/networking/params.pp
sebastian-miclea: I had a second look on what it does with these vars and updated.
It seems that none of them are required, in the networking.pp manifest it just tries to comment a line (which doesn't exist in Fedora 29). Similar thing for the udev vars, $udev_rule gets deleted, and $udev_rule_gen gets symlinked to /dev/null. In Fedora 29 it's already in this state so I've set everything to 'undef'

sebastian-miclea:
Imaging pipeline passed on dev:

https://jenkins-sre.delivery.puppetlabs.net/view/__experimental%20automatic/job/experimental_auto_fedora-29_x86_64_vmware_vsphere.nocm_test_template_on_vmpooler_dev_packer/

sebastian-miclea:
Dependent on: https://github.com/puppetlabs/ci-job-configs/pull/5322

sebastian-miclea:
https://github.com/puppetlabs/ci-job-configs/pull/5322 Has been merged.

PR #388, opened 2018-11-23
(IMAGES-840) WinRm https access

jcoconnor:
Enable HTTPS Winrm access during the vmpooler post-clone operation where we know what the machines DNS name is going to be.

Win-7/2008/2008r2 requires the makecert utility, whereas the later versions of windows can use powershell cmdlets for the certificate generation.


templates/win/common/scripts/vmpooler/vmpooler-post-clone-configuration.ps1
```diff
@@ -96,6 +95,14 @@ $CygwinGroupFile = "$CygwinDir\etc\group"
 Invoke-Expression $CygwinMkpasswd | Out-File $CygwinPasswdFile -Force -Encoding "ASCII"
 Invoke-Expression $CygwinMkgroup | Out-File $CygwinGroupFile -Force -Encoding "ASCII"
 
+# Winrm over HTTPS Configuration
+Write-Output "Configurating WinRM over HTTPS"
+# Generate Self-Signed Cert with DNS name
+$DNSHostname = "$NewVMName.delivery.puppetlabs.net"
```
Iristyle: Given the script name `vmpooler-post-clone-configuration.ps1`, I'm assuming this runs as part of booting the VM after it's acquired a new hostname?

jcoconnor: Correct - though if you look down the script, the host rename happens at the very end.

templates/win/common/scripts/vmpooler/vmpooler-post-clone-configuration.ps1
```diff
@@ -96,6 +95,14 @@ $CygwinGroupFile = "$CygwinDir\etc\group"
...
+$NewCert = New-SelfSignedCertificate -DnsName "$DNSHostname" -CertStoreLocation Cert:\LocalMachine\My
```
Iristyle: For reference, Bolt tests use a "well known" cert bound to `localhost` - I think what you've done here is necessary to allow incoming connections without additionally asking clients to ignore the hostname mismatch.

For reference, the Bolt script does this to import the cert:

```powershell
($cert = Import-PfxCertificate -FilePath resources/cert.pfx -CertStoreLocation cert:\\LocalMachine\\My -Password (ConvertTo-SecureString -String bolt -Force -AsPlainText)) | Format-List
```
With the code at https://github.com/puppetlabs/bolt/blob/master/appveyor.yml#L118 using the pre-generated `cert.pfx` at https://github.com/puppetlabs/bolt/blob/master/resources/cert.pfx

But as mentioned the hostname is `localhost` per:

```powershell
New-WSManInstance -ResourceURI winrm/config/Listener -SelectorSet @{Address='*';Transport='HTTPS'} -ValueSet @{Hostname='localhost';CertificateThumbprint=$cert.Thumbprint} | Format-List
```

jcoconnor: Thanks @Iristyle - Yes this code is referenced in the ticket and I used it as a starting place.
However, I decided it was probably better to go for the Self Cert based on the DNS name - as you point out - it avoids the need to ignore the host name mismatch.

jcoconnor:
@Iristyle This is revised to use ```makecert``` for the legacy versions (win-2008/2008r2/7).

I'm running test builds for these at the moment - but should be able to confirm soon if the builds succeed.

jcoconnor:
Testing on Win-7/2008r2 showed that the WinRM/HTTPS enabling has to be done in the ```startup``` script and not the ```post-clone``` stage as the hostname has to be renamed before it will register.

Was able to get Remoting/https working manually but need to run this through an initial sample, followed by a full run to verify that it works consistently across all versions.

jcoconnor:
Hmm  - not quite there - this is a victim of the Delayed Startup on the WinRm service.
So need to wait until service is started before doing the configure.

Added a wait check.
Going to build all the instances overnight, but think this PR is ready for merge.
/Ping @Iristyle @michaeltlombardi

templates/win/common/scripts/vmpooler/vmpooler-clone-startup.ps1
```diff
@@ -1,38 +1,63 @@
...
-
-#--- SCRIPT ---#
-Write-Output "Updating the Cygwin passwd file!"
-
-#Update the passwd file.
-Invoke-Expression $CygwinMkpasswd | Out-File $CygwinPasswdFile -Force -Encoding "ASCII"
-Invoke-Expression $CygwinMkgroup | Out-File $CygwinGroupFile -Force -Encoding "ASCII"
-
-Write-Output "Bye"
+# VMPooler Startup Script
```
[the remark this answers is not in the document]
Iristyle: Did you mean that the original was LF?  We can keep the CRLF and just add to .gitignore...

templates/win/common/scripts/vmpooler/vmpooler-post-clone-configuration.ps1
```diff
@@ -81,7 +81,6 @@ Write-Output "Register the Cygwin LSA authentication package "
 # Add github.com as a known host (needed for git@gihub:<repo> clone ops)
 & $CygWinShell --login -c `'ssh-keyscan -t rsa github.com `>`> /home/$AdminUsername/.ssh/known_hosts`'
 
-
```
Iristyle: Unnecessary change

templates/win/common/scripts/provisioners/install-win-packages.ps1
```diff
@@ -65,5 +65,11 @@ Set-ItemProperty -Path $RegPath -Name PATH -Value $NewPath
 
 Write-Output "Sysinternal Tools Installed"
 
+if ($WindowsVersion -Like $WindowsServer2008 -or $WindowsVersion -Like $WindowsServer2008R2)
+{
+  Write-Output "Grabbing Makecert for the Legacy versions among us"
+  Download-File "https://artifactory.delivery.puppetlabs.net/artifactory/generic/buildsources/windows/makecert/makecert.exe" "$PackerScripts\makecert.exe"
```
Iristyle: Would be nice to remove this an external dependency if possible.... doesn't it ship in the box already? Can we embed the binary in the repo (not sure how large it is)?

templates/win/common/scripts/vmpooler/vmpooler-clone-startup.ps1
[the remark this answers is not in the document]
jcoconnor: Ok - could do that - generally I have been moving all the scripts to LF.

templates/win/common/scripts/vmpooler/vmpooler-clone-startup.ps1
```diff
@@ -1,38 +1,68 @@
...
+    {
+        # Windows 7/2008/2008R2 as always need compatibility code.
+        # Using instructions from: https://blog.jayway.com/2011/11/21/winrm-w-self-signed-certificate-in-4-steps/
+        # A copy of makecert was obtained from: https://stackoverflow.com/questions/5510063/makecert-exe-missing-in-windows-7-how-to-get-it-and-use-it
+        Write-Output "Generating Certification (makecert)"
+        & $PackerScripts/makecert -sk "$DNSHostname" -ss My -sr LocalMachine -r -n "CN=$DNSHostname" -a sha1 -eku "1.3.6.1.5.5.7.3.1"
+        $global:NewCert = ls cert:LocalMachine\my
+    }
+    else {
+        # Otherwise use the cmdlet provided by Powershell to generate the certificate.
```
Iristyle: Can we remove this branch of code and just use `makecert` everywhere?

templates/win/common/scripts/vmpooler/vmpooler-clone-startup.ps1
```diff
@@ -1,38 +1,68 @@
...
+        Write-Output "Generating Cert (New-SelfSignedCertificate)"
+        $global:NewCert = New-SelfSignedCertificate -DnsName "$DNSHostname" -CertStoreLocation Cert:\LocalMachine\My
+    }
+    # Enable HTTPS - once the WinRm service is actually up.
+    Write-Output "Checking/Waiting for running WinRm"
+    $WinRmService = Get-Service -Name WinRM
+    $WinRmService.WaitForStatus("Running")
```
Iristyle: Probably want a timeout?

templates/win/common/scripts/vmpooler/vmpooler-post-clone-configuration.ps1
jcoconnor: ok - will note and remove in a later ticket unless you think I should drop it now.

templates/win/common/scripts/provisioners/install-win-packages.ps1
jcoconnor: Yeah - can add it to the Repo instead - I've generally put .exe's in artifactory tho.
Its pretty small.

templates/win/common/scripts/vmpooler/vmpooler-clone-startup.ps1
jcoconnor: Yes - will do that tomorrow.

templates/win/common/scripts/vmpooler/vmpooler-clone-startup.ps1
jcoconnor: Ok - good idea - mind you - if WinRm isn't going to come up, there is a problem.......

templates/win/common/scripts/vmpooler/vmpooler-post-clone-configuration.ps1
jcoconnor: Sorted 👍

templates/win/common/scripts/provisioners/install-win-packages.ps1
jcoconnor: Sorted 👍

templates/win/common/scripts/vmpooler/vmpooler-clone-startup.ps1
jcoconnor: Sorted 👍

templates/win/common/scripts/vmpooler/vmpooler-clone-startup.ps1
jcoconnor: Sorted 👍

jcoconnor:
@Iristyle Pushed up again - with changes added including putting makecert into the repo.

jcoconnor:
I have tested this across all the windows builds from ```win-2008-x86_64``` right through to ```win-2019-x86_64``` (and the desktop variants) and am able to WinRM/ssh into all of the versions.

The only anomalies were:
* win-10-1511-x86_64
* win-81-x86_64

Both of these need the credential username to be "LOCAL\Administrator" rather than just "Administrator"

So this PR is ready to go.

jcoconnor:
Ping @Iristyle @michaeltlombardi

jcoconnor:
I have added a commit to update the version for the next images refresh.

templates/win/common/scripts/vmpooler/vmpooler-clone-startup.ps1
```diff
@@ -35,4 +35,32 @@ Write-Output "Updating the Cygwin passwd file!"
...
 Invoke-Expression $CygwinMkgroup | Out-File $CygwinGroupFile -Force -Encoding "ASCII"
 
+if (-not (Test-Path "$PackerLogs\WinRMHTTPS.installed"))
+{
+    # First wait until WinRm Service is available - allowing a very generous startup time
+    # Seems to be needed as 30 seconds timed out on some machines.
+    Write-Output "Checking/Waiting for running WinRm"
+    $WinRmService = Get-Service -Name WinRM
+    $WinRmService.WaitForStatus("Running",'00:02:00')
+    Start-Sleep -Seconds 2
```
Iristyle: Wait up to 2 minutes for service.. then an extra 2 seconds?

jcoconnor: Added comment as suggested - will look at removing when doing follow up work.

PR #390, opened 2018-11-30
(IMAGES-896) Remove Build Hostname

jcoconnor:
This is associated with changes in platform-ci-utils where the
host name is no longer needed, or can default to blank.


jcoconnor:
This PR is needed to support https://github.com/puppetlabs/platform-ci-utils/pull/46

Have just tested this on ```win-10-ent-i386``` and it appears good to go.

PR #393, opened 2018-12-10
(IMAGES-941) Create machine image(s) for macOS 10.14 Mojave and fix rc local for osx templates

loopinu:
fixed missing semicolon in vsphere/rc.local  & added execute permissions in vsphere/params.pp


loopinu:
Hi Enis, i added on this PR also macos 10.14 because i had to test this to see if it works correctly. I created the template and added to vmpooler. There is still a problem with the boot_command it's not yet finished because it get's stuck at some point at the installer. Untill i manage to fix it i'll have the do not merge tag on this PR.
I will make another PR that fixes the rc local because it will impact other imaging templates

PR #397, opened 2019-01-02
(maint) Set tools_upload_flavor to linux in top-level vmware.base.json

ekinanp:
Without this variable, any Linux images built from the top-level
vmware.base.json template will error in the vmtools setup step because
the linux.iso file will not be uploaded onto the image.


ekinanp:
@underscorgan This should fix the build issues you encountered in the imaging pipelines.

PR #398, opened 2019-01-03
(IMAGES-851) Stop Notepad++ update check

jcoconnor:
Disable the Notepad++ updater during the installation to suppress the irritating update prompt when running the editor.
Version 7.5.2+ of npp conveniently provides an install option (/noUpdater) for this.

Also take the opportunity to puppetise the installation of this package and update the version to 7.6.2

The windows package install tests need to be moved to the "puppet" test stage as the Notepad++ is no longer installed by the "install-win-packages.ps1" provisioner. This provisioner will be removed once the remainder of the windows packages are puppetised. (see IMAGES-636)


templates/win/common/puppet/windows_template/manifests/apps/notepadplusplus.pp
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  download_file { "${notepadppinstaller}" :
+    url                   => "${notepadppdownloadurl}/${notepadppinstaller}",
+    destination_directory => $::packer_downloads
+  }
+  -> package { "Notepad++ ${notepadpparchtype}":
+    ensure          => installed,
+    source          => "${::packer_downloads}\\${notepadppinstaller}",
+    install_options => ['/S','/noUpdater']
```
Iristyle: So the important change is the `/noUpdater` then, correct?

Iristyle: Added in 7.5 per https://notepad-plus-plus.org/download/v7.5.html

👍 	from me

jcoconnor: Yes - I just found that accidentally after perusing a lot of other suggestions on how to fix it.

But since I've had a long term goal to puppetize the package installs it seemed to be an opportunity to try that too.

templates/win/common/puppet/windows_template/manifests/apps/notepadplusplus.pp
```diff
@@ -0,0 +1,26 @@
+# Installs and configures NotePad Plus Plus
+# Sets the Do Not check for update flag once installed.
+
+class windows_template::apps::notepadplusplus()
+{
+  $notepadppdownloadurl = 'https://notepad-plus-plus.org/repository/7.x/7.6.2'
```
Iristyle: Do we care that we're also switching from Artifactory to an upstream URL? Does this increase chances of failed builds at all?

jcoconnor:
Marking don't merge until I get a clean build thru.

templates/win/common/puppet/windows_template/manifests/apps/notepadplusplus.pp
jcoconnor: I've reverted back to artifactory for the moment.
There is a separate ticket to make this build more accessible to the community by either moving to direct site download, or making it configurable, so leaving this for another day.

jcoconnor:
Commit comments updated so this is ready for merge pending a run on a sample of builds including the "awkward squad".

templates/win/common/puppet/windows_template/manifests/apps/notepadplusplus.pp
jcoconnor: Have updated commit and comments internally to capture this point.

PR #401, opened 2019-01-08
(RE-11841) Update RHEL6 mirrors

underscorgan:
This updates the RHEL 6 images to use the artifactory RHEL mirrors.

This also removes configuration of the `lb` repo as our subscription seems to not have access to that. Working to figure out if this is OK for haproxy module testing. If we need to re-add it it can probably happen in a separate PR, the existing cache of the lb repo seems to have very recently expired/corrupted so we can't install any packages from there currently anyways.


underscorgan:
Built dev images and they seem to be able to pull from the new mirrors 🎉

PR #402, opened 2019-01-09
(IMAGES-952) Add Powershell 6

jcoconnor:
Add Powershell 6 (Core) to the Windows Templates.

WMF 4.0 is a minimum requirement for the Powershell Core installer, so suppress the installation for windows platforms with WMF 2.0 and 3.0

### DIFF
--- a/templates/win/common/puppet/windows_template/manifests/apps/powershell6.pp
+++ b/templates/win/common/puppet/windows_template/manifests/apps/powershell6.pp
@@ -1,0 +1,25 @@
+# Installs and configures Powershell 6 (core)
+
+class windows_template::apps::powershell6()
+{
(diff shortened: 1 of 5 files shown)


jcoconnor:
This builds and operates fine as long as WMF Version is 4.0 or greater, i.e. it fails on Windows 2008, 2008r2 and 2012.

Think the best approach here is to not to attempt to install it on the earlier platforms.

jcoconnor:
Added check to silently *not* install if WMF < 4

jcoconnor:
This checks out ok on build for:
* win-2012r2-x86_64
* win-2016-x86_64
* win-10-ent-x86_64
* win-10-ent-i386
* win-2008r2-x86_64 (doesn't install and pester testing doesn't check for it)
* win-2008r2-wmf5-x86_64 - Installs ok for this as it is WMF 5

So this is ready for merge.

PR #403, opened 2019-01-10
(IMAGES-1000) Update VM Hardware Type

jcoconnor:
Moving vm_version to 14 (Esx 6.7) as all the VMWare hosts are at
this hardware level.


jcoconnor:
Have verified that VM's now build and boot up HW Type 14 "ESXi 6.7 and later (VM version 14) "

PR #404, opened 2019-01-10
(IMAGES-636) Move Win Pkg installs to Puppet stage

jcoconnor:
Puppetise the installation of the following packages:
1. Chrome
2. Git for Windows
3. Sysinternals

This means that the install-win-package provisioner can be removed.


jcoconnor:
This is ready for review - running it through a build test first.

jcoconnor:
I have run a build test on ```win-2019-x86_64``` and verified that:
1. Chrome is installed
2. Git for windows is installed.
3. Sysinternals utilities are installed.
4. PATH variable is updated correctly and survives a reboot.
5. The win-packages and sysinternals Pester tests pass ok.

So this is good to merge.

PR #408, opened 2019-01-14
(IMAGES-1017) Localise TZ Test

jcoconnor:
French uses a different timezone title in the systeminfo output so add
localisation support to enable Pester tests to work correctly.


jcoconnor:
French Build has completed successfully (at last), so this is ready for merge.

PR #412, opened 2019-01-16
Split out yum repos for rhel 7 for yum group support

shrug:
Artifactory does not merge yum groups in a virtual repo (https://www.jfrog.com/jira/browse/RTFACT-14058), this PR breaks out the os, extras and optional repos to individual yum configs so the systems can use yum groups.


ScottGarman:
:+1:

ScottGarman:
Actually, is it possible to remove the repos manifest section checking for major release 7 and have that fall back to the default?

shrug:
I didn't see that, looks like it should work though yes. Updated.

PR #418, opened 2019-01-25
(IMAGES-1005) Install Platform9 Drivers

jcoconnor:
Install the necessary drivers for Platform 9 including:
1. Download .iso and install drivers
2. Install Cloudbase-Init and qme agent
3. Setup Post-Install to run cloudbase-init and modify sysprep actions appropriately.
4. Disable VMWare services (a later ticket will investigate if they can be removed completely).
5. Some additional cleanup/removal of unnecessary files.


jcoconnor:
This is ready for review - I need to run it through a final set of tests and will note the test results as they are carried out.

templates/win/common/files/PostCloneTemplate.xml
```diff
@@ -106,6 +106,27 @@
...
+                    <CommandLine>reg.exe ADD "HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Policies\System" /v "EnableLUA" /t REG_DWORD /d 0 /f</CommandLine>
+                    <Description>Disable UAC</Description>
+                    <Order>1</Order>
+                </SynchronousCommand>
+                <SynchronousCommand wcm:action="add">
+                    <CommandLine>cmd /C start /wait NET ACCOUNTS /MAXPWAGE:UNLIMITED</CommandLine>
+                    <Order>2</Order>
+                </SynchronousCommand>
+                <SynchronousCommand wcm:action="add">
+                    <CommandLine>cmd.exe /c wmic useraccount where &quot;name=&apos;__ADMIN_USERNAME__&apos;&quot; set PasswordExpires=FALSE</CommandLine>
```
denis-susan: is special character escaping necessary? if yes why is the disable uac command not escaped?

templates/win/common/scripts/common/init-sysprep.ps1
```diff
@@ -35,19 +39,33 @@ if (Test-Path "$PackerLogs\AppsPackageRemove.Required") {
...
 
 # Stop the tilemode service
 Stop-Service -Name "tiledatamodelsvc" -Force -Verbose -ErrorAction SilentlyContinue
 Stop-Service -Name "AppXSvc" -Force -Verbose -ErrorAction SilentlyContinue
 Stop-Service -Name "staterepository" -Force -Verbose -ErrorAction SilentlyContinue
 
+if ($RemoveVMWare) {
+    # Disable VMWare tools - would prefer to uninstall but haven't been able to get the following command to work reliably.
+    # Uninstall VMWare Tools using the GUID obtained from HKLM\Software\Microsoft\Windows\CurrentVersion\Uninstall
+    # Start-Process "msiexec" -Wait -NoNewWindow -Verbose -ArgumentList "/passive /x {F32C4E7B-2BF8-4788-8408-824C6896E1BB} /norestart REBOOT=REALLYSUPPRESS"
```
denis-susan: we could try removing it from powershell native using this:
``` 
(gwmi Win32_Product | ? Name -eq "VMware Tools").uninstall()
```

templates/win/common/files/PostCloneTemplate.xml
jcoconnor: Yes - character escaping is essential here since we are using XML.
Good catch on the EnableLUA - think that was missed in a previous edit, i.e. Line 71 doesn't have escaping whereas line 92 has.

Fixing.....

templates/win/common/scripts/common/init-sysprep.ps1
jcoconnor: Will give that a go later - but not sure this method allows us to provide options (e.g. /quiet).
Think we might be hitting some interaction with the OS here - if I run these commands manually on the VM, they work fine, but in a script, the uninstall never returns, i.e. procedure hangs.

jcoconnor:
So - I have successfully built the image, manually converted and imported it to platform9.
It starts and runs ok, so this PR is ready for merge.

templates/win/common/scripts/common/init-sysprep.ps1
```diff
@@ -35,19 +39,33 @@ if (Test-Path "$PackerLogs\AppsPackageRemove.Required") {
...
+    Write-Output "Disabling VMWare Tools"
+    Set-Service -Name "VGAuthService" -StartupType Disabled -Verbose -ErrorAction SilentlyContinue
+    Set-Service -Name "VMwareCAFCommAmqpListener" -StartupType Disabled -Verbose -ErrorAction SilentlyContinue
+    Set-Service -Name "VMwareCAFManagementAgentHost" -StartupType Disabled -Verbose -ErrorAction SilentlyContinue
+    Set-Service -Name "VMware Physical Disk Helper Service" -StartupType Disabled -Verbose -ErrorAction SilentlyContinue
+    Set-Service -Name "vmvss" -StartupType Disabled -Verbose -ErrorAction SilentlyContinue
+    Set-Service -Name "VMTools" -StartupType Disabled -Verbose -ErrorAction SilentlyContinue
```
RandomNoun7: This block might be easier to read and manage if you stick all of the services you want to modify into an array so you can look at the list of services separately from the code that acts on the services like so:

```powershell
    $services = @(
      "VGAuthService"
      "VMwareCAFCommAmqpListener"
      "VMwareCAFManagementAgentHost"
      "VMware Physical Disk Helper Service"
      "vmvss"
      "VMTools"
    )

    Get-Service -Name $services | 
      Set-Service -StartupType Disabled -Verbose -ErrorAction SilentlyContinue
```

RandomNoun7: If that syntax gives you issues on Powershell 2 just put commas after each service name and that will likely clear it up.

templates/win/common/scripts/common/init-sysprep.ps1
```diff
@@ -35,19 +39,33 @@ if (Test-Path "$PackerLogs\AppsPackageRemove.Required") {
...
-if ( ($WindowsVersion -like $WindowsServer2008R2) -or ($WindowsVersion -like $WindowsServer2008) ) {
+$SysPrepArgs = "/generalize /oobe /quit /unattend:C:\Packer\Config\post-clone.autounattend.xml"
+if ( ($ImageProvisioner -eq "vmware") -and ($WindowsVersion -notlike $WindowsServer2008R2) -and ($WindowsVersion -notlike $WindowsServer2008)) {
     # /mode:vm was only introduced from win-2012 onwards
-    $SysPrepArgs = "/generalize /oobe /quit /unattend:C:\Packer\Config\post-clone.autounattend.xml"
+    Write-Output "Using /mode:vm"
+    $SysPrepArgs += "/mode:vm"
 }
 Write-Output "Starting the Sysprep Process"
 $zproc = Start-Process "$SysPrepDir\sysprep.exe" @SprocParms -ArgumentList "$SysPrepArgs"
```
RandomNoun7: I see a variable here called SprocParms being used as a splatter, but I don't see where that variable is ever declared or assigned with a value in this file. Is it coming from some other scope I don't see?

templates/win/common/scripts/common/init-sysprep.ps1
jcoconnor: Ok - after discussion with @denis-susan Think this might be related to the ```winrm``` session dropping when ```vmware tools``` are removed, as the network connection getting clobbered.

Have raised a new ticket for this to deal with seperately.

templates/win/common/scripts/common/init-sysprep.ps1
jcoconnor: Thanks - will do that if I have no success in addressing the vmware tools removal (ticketed seperately).

templates/win/common/scripts/common/init-sysprep.ps1
jcoconnor: Yip - defined in ```windows_env.ps1``` - its used in several places.

jcoconnor:
Build runs cleanly to the end with the latest fix to the vmware tools problem.
Will verify that the image is ok tomorrow before we can hit the ⏭ on this.

jcoconnor:
Have verified that the image initialises and boots up correctly on Platform9, so this PR is ready for merge.

PR #426, opened 2019-02-19
(IMAGES-1025) Add debian-10.0-x86_64 template

gimmyxd:
build debian 10 on vcenter

### DIFF
--- a/manifests/modules/packer/manifests/vsphere/params.pp
+++ b/manifests/modules/packer/manifests/vsphere/params.pp
@@ -1,6 +1,6 @@
 # == Class: packer::vsphere::params
 #
-# Default vsphere parameter values for the packer module 
+# Default vsphere parameter values for the packer module
 #
 class packer::vsphere::params {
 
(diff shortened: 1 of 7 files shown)


manifests/modules/packer/manifests/vsphere/params.pp
```diff
@@ -69,10 +69,10 @@
         $periodic_file         = '/etc/apt/apt.conf.d/10periodic'
       }
       $ruby_package          = [ 'ruby' ]
-      $repo_name             = 'debian'
+      $repo_name             = 'debian__remote'
       $repo_list             = 'main contrib non-free'
-      $security_repo_name    = 'debian-security'
-      $security_release      = "${facts['lsbdistcodename']}/updates"
+      $security_repo_name    = 'debian__remote'
+      $security_release      = "${facts['lsbdistcodename']}"
```
branan: It looks like the way this was done here actually disabled the security updates repo entirely, instead making it a clone of the base packages configuration.

@mwaggett is setting up the security repo in artifactory and we'll get this updated to point to the right place.

PR #430, opened 2019-03-06
(Images 1067)-Change MacOS 10.12 packer template to use vcenter build

loopinu:
modified macos 10.12 template to build on vcenter


loopinu:
Closing this due to drop of support for osx 10.12

PR #435, opened 2019-03-14
(IMAGES-1018) Puppet Errors aren't checked correctly

jcoconnor:
Revise the puppet Configure stage to run as a scheduled task so that
need for rebooting is properly deteched and acted on.

Within this, the configure loop to properly trap and address Puppet
configuration errors - to date these have been logged but not
correctly actioned.


jcoconnor:
I've run tests builds using the Mock 2019 harness on both Platform9 and vcenter builds - the build dynamic data is picked up correctly by Puppet and written to the relevant permanent ENV variables.

Have also been able to force puppet errors and demonstrate that they are caught correctly and logged.

jcoconnor:
Do Not Merge added - two issues discovered:
1. Need to put in flag and detect into test suite to ensure puppet failure stops build.
2. Win 2008 (only) needs to print out puppet log.

jcoconnor:
So - the Puppet succeeded test is added and working.
Waiting for a win-2008 build to complete to verify that (takes ages)

jcoconnor:
Win-2008 (non-R2) builds ok.

So this PR is ready for review/merge.

templates/win/common/scripts/common/puppet-configure.ps1
```diff
@@ -0,0 +1,144 @@
...
+& reg unload HKLM\DEFUSER
+
+
+# Puppet Configure Cycle complete - delete this task.
+Write-Output "Deleting Bootstrap Scheduled Task"
+schtasks /Delete /tn PuppetConfigure /F
+
+# Enable WinRM so Packer control will resume after reboot.
+Set-Service "WinRM" -StartupType Automatic
+Write-Output "WinRM Enabled - Packer will resume next reboot"
```
RandomNoun7: If Set-Service is not in a try catch block, are we sure this will be true?

templates/win/common/scripts/common/windows-env.ps1
```diff
@@ -143,6 +145,7 @@ Function Create-PackerStagingDirectories {
   if (-not (Test-Path "$PackerStaging\puppet\modules")) {
     Write-Output "Creating $PackerStaging"
     mkdir -Path $PackerStaging\puppet\modules
+    mkdir -Path $PackerStaging\puppet\data
```
RandomNoun7: This path here is the same as the path constructed for $PuppetModulesPath on line 96 above. This means that if you change the path on 96, but forget to change it here you will have a mismatch. Is that likely? I don't think this is likely to change, but why type the same path twice?

jcoconnor: Good point - I've rewritten this to:
1. Replace all paths with their respective variables.
2. Gone all pedantic and used New-Item -ItemType Directory with -Force
3. Used the Touch-File model to avoid repetition as with other blocks.

templates/win/common/scripts/common/puppet-configure.ps1
jcoconnor: In this case I think I can just turn WinRM as well as enabling the service.
But you are correct - not sure at all what we can do if this fails other than the restart operation will eventually time out - but the log then gets lost on the destroyed machine.

Sort of catch-22 ?

templates/win/common/scripts/common/windows-env.ps1
jcoconnor: This worked nicely once I tested it :)

templates/win/common/scripts/common/puppet-configure.ps1
jcoconnor: I've re-modelled this in the code

templates/win/2008/x86_64/vmware.vcenter.cygwin.json
```diff
@@ -39,11 +43,12 @@
 
   "builders": [
     {
-      "type": "vsphere-iso",
+      "type": "vsphere-23-iso",
```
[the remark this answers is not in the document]
jcoconnor: Yes - lets ticket that just to make sure we do all the fixups.

PR #436, opened 2019-03-21
(IMAGES-976) Improve build parameter injection

jcoconnor:
Continue the work from IMAGES-1018 to remove environment variables to
inject data into the build:

1. Switch to .json format as this is easier to consume in powershell. This is done with PS2 support for the Windows-2008/2008r2 platform.
2. This requires changes to the puppet lookup and the variables used to access the .json data.
3. Windows Version/Edition information is passed in via .json file to be consumed by the packer testing.
4. This required moving the platform-package file processing to the bootstrap stage mainly for Windows-2008/2008r2 since .Net 3.5 is required to parse the .json format.

Note - this change has quite a bit of copy/pasta as the packer provider to write the .json file is repeated in several packer scripts. It is hoped to compress most of the windows build into a single packer script in a later ticket/PR which will eliminate much of the duplication.

### DIFF
--- a/templates/win/2008/x86_64/vmware.vcenter.cygwin.json
+++ b/templates/win/2008/x86_64/vmware.vcenter.cygwin.json
@@ -36,7 +36,9 @@
     "boot_wait"                 : "10ms",
     "convert_to_template"       : "false",
     "network_card"              : "vmxnet3",
-    "boot_order"                : "disk,cdrom"
+    "boot_order"                : "disk,cdrom",
+
+    "valid_exit_codes"          : "0"
   },
 
   "description": "Builds a Windows Server 2008 SP2 template VM for use in VMware. Custom spec needed for this platform",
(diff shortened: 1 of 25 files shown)


jcoconnor:
Just a bit of background on this - @jpogran & @michaeltlombardi have convinced me I'm an idiot to be trying to parse ```yaml``` in powershell - so looking to see if I can use ```json``` as a backend instead.

Watch this space.....

jcoconnor:
I've put this thru tests on a sample of versions including:
* Windows-2008 (some rework of the .Net 3.5 install sequence was needed here)
* Windows-2008r2 (same issue as above)
* Windows-2012r2
* Windows 2019
Once the usual initial glitches were resolved with the earlier versions of windows, all appear to build ok and start up/run the acceptance test suite.

Comments appreciated.....

PR #437, opened 2019-03-27
(IMAGES-1082) update to PSWindowsUpdate 2.1.1.2

jcoconnor:
1. Update to Version 2.1.1.2 of the PsWindows Update tool as this appears to run faster for the Windows 10 series of builds. Retain earlier version for win-2012r2 due to WUHistory problem.
2. Add in specific patches to address Windows Service Stack (Update) issues that cause problems for the Windows 2016 build. This hopefully shortens the update time for Windows 2016 by avoding two cumulative update downloads and random failures to do the windows update.
3. Windows 10/2016 now appears to work correctly (and faster) with WSUS so enable these. Windows 10 2016 LTSB still fails to download from WSUS and Win-2012 doesn't get any updates from WSUS, so continue to exclude these.


jcoconnor:
@denis-susan & @sebastian-miclea This is ready for review.

Please do not squash these commits in case I need to reverse individual changes.
Tx

PR #438, opened 2019-03-28
(IMAGES-1084) Do not deploy updates repo on Jessie

branan:
As part of the LTS process (apparently?), Debian has removed
jessie-updates in favor of the security updates repo.

This commit updates the repos configuration to only deploy the
-updates repo for Debians newer than Jessie. It also fixes the
security updates repo to ensure we can still get updated versions of
packages, rather than only the Jessie base system.

### DIFF
--- a/manifests/modules/packer/manifests/vsphere/repos.pp
+++ b/manifests/modules/packer/manifests/vsphere/repos.pp
@@ -51,14 +51,16 @@ class packer::vsphere::repos(
         },
       }
 
-      apt::source { "${facts['lsbdistcodename']}-updates":
-        release  => $updates_release,
-        location => "${repo_mirror}/${repo_name}",
-        repos    => $repo_list,
-        include  => {
-          'src' => true,
-          'deb' => true,
-        },
+      if $facts['operatingsystem'] == 'Debian' and $facts['operatingsystemrelease'] >= 9 {
+        apt::source { "${facts['lsbdistcodename']}-updates":
+          release  => $updates_release,
+          location => "${repo_mirror}/${repo_name}",
+          repos    => $repo_list,
+          include  => {
+            'src' => true,
+            'deb' => true,
+          },
+        }
       }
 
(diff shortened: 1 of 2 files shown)


branan:
Unfortunately there are no unit tests for the puppet code in this repo - it passed `puppet parser validate` on my laptop, but please look closely for typos in fact names/values

manifests/modules/packer/manifests/vsphere/repos.pp
```diff
@@ -51,14 +51,16 @@
...
 
-      apt::source { "${facts['lsbdistcodename']}-updates":
-        release  => $updates_release,
-        location => "${repo_mirror}/${repo_name}",
-        repos    => $repo_list,
-        include  => {
-          'src' => true,
-          'deb' => true,
-        },
+      if $facts['operatingsystem'] == 'Debian' and $facts['operatingsystemrelease'] >= 9 {
```
underscorgan: This isn't working on my test box, I'm getting:
```
Error: Evaluation Error: Comparison of: String < Integer, is not possible. Caused by 'A String is not comparable to a non String'. at /root/test.pp:1:82 on node jblxprmdqygwb7h.delivery.puppetlabs.net
```

I was able to get it working as expected with 
```
if $facts['operatingsystem'] == 'Debian' and $facts['operatingsystemmajrelease'] =~  /^(7|8)/
```

branan: I'll take a look in the morning. I had no way of testing this on my laptop other than a simple `parser validate`.

If the release is a string, the regex seems best

underscorgan: yeah, I tested it on a jessie box with as old as agent 1.10.x and as new as 6.latest and it was an integer for all of them.

branan:
@mihaibuzgau has an updated version of this on his branch with the rest of the Jessie work. I'm gonna go ahead and close this in favor of his final PR

PR #442, opened 2019-04-02
(IMAGES-1082) Fix WinRM Upload errors

jcoconnor:
This PR fixes a number of upload issues that were observed when testing IMAGES-1082 across the full range of builds.

This turned out to be due to sequencing errors introduced with both this and the previous ticket IMAGES-976 where json parameter files were used to inject build data.

The Windows "Desktop Feature" which is enabled in the Server editions appears to be the main culprit as it interacts with the WMF Enable sequence which allows the packer file uploads to start before the initial sequence is complete - causing a reboot mid-upload which aborts the build.

So the original order of the platform-package.ps1 processing has been reverted to happen during the windows-update processing and the .Net 3.5 enabling which is required for .json processing is brought forward to the bootstrap phase for Windows 2008/2008r2 only.

Another sequencing weaknesses was also tied down including disabling WMF before the network is connected (i.e. Hypervisor tools installed).

A second commit to this PR addresses IMAGES-1095 by adding the "AddTrust External CA Root" cert for Windows 7/2008/2008r2 builds.


jcoconnor:
This PR replaces #440

jcoconnor:
This appears to fix the Windows-2008/2008r2 build errors, but running it through all the builds to be certain.

jcoconnor:
This has been tested on all Windows Builds and appears to resolve the sequencing issues.
There is one remaining issue with the ```win-7-x86_64``` build which hangs on the puppet run.

Ticketing this for separate investigation.

jcoconnor:
Putting a don't merge on this as going to address the SSL/artifactory/win7 issue.

jcoconnor:
Added additional commit to address IMAGES-1095 where it wasn't able to download from artifactory on earlier versions of Windows.

Running this through test on the following builds to verify:
* win-2008-x86_64
* win-2008r2-x86_64
* win-2008r2-wmf5-x86_64
* win-7-x86_64
* win-7-wmf5-x86_64

PR #449, opened 2019-04-16
(IMAGES-1098) Update slipstream builds

jcoconnor:
The Slipstream template hasn't followed recent updates in the common templates, so give it some TLC to bring it up to spec and to get it work on the vCenter builds.

Also include the boot image .iso name in the json manifest placed on the image.

The changes are also "copied" to Windows 2008, although this image needs some additional manual work to generate and finalise the image as DISM doesn't appear to work satisfactorally in the platform. This platform is near EOL so this is expected to be the last Slipstream update for it, but is still considered worth it as it embeds 95 updates in the .iso

The .Net Framework was also updated to .Net 4.7.2 as this significantly reduces the number of .Net related updates.


jcoconnor:
Running this through test at the moment - looks good for most of the platforms - but need to test the actual .iso files by attempting to build off them.

jcoconnor:
Blocking to update the builds with the new slipstreams.

jcoconnor:
I've completed build/test on this for the following builds:
- win-2008-x86_64
- win-2008r2-x86_64
- win-2008r2-wmf5-x86_64
- win-2012-x86_64
- win-2012r2-x86_64
- win-2012r2-wmf5-x86_64
- win-2012r2-core-x86_64
- win-7-x86_64
- win-81-x86_64
- win-2016-x86_64 (unchanged - just checking)
- win-2016-core-x86_64 (unchanged - just checking)

All build showing significant reduction in the number of updates installed - in most cases its .Net updates that can't be slipstreamed (although this issue seems to be fixed for later versions of .Net).

There are two issues:
1. win-2012r2-core fails windows update testing as the .iso contains all the updates and therefore there are no updates to be installed - have relaxed the Pester testing on this until next month when the May updates will make this problem go away.
2. win-2012-x86_64 - this doesn't built with WSUS so wasn't able to make significant updates to the slipstream.

May look at the win-2012-x86_64 build later in a separate ticket but given the significant improvement to the other builds, this is ready to go.

PR #450, opened 2019-04-23
(IMAGES-1100) Enable Netbios/lmhosts

jcoconnor:
Netbios/lmhosts was disabled pre-sysprep but not re-enabled which causes
issues setting up AD etc.

So set services back to default enabled.


jcoconnor:
@Iristyle @glennsarti Can you merge this if you want it.
I'm due to redeply the Platform9 images this week but can include this in the deploy if you let me know today.

Thanks - John

PR #451, opened 2019-05-08
(IMAGES-1101)- fix debian 8 ip address reset after reboot

loopinu:
This PR fixes debian 8 retrieving an ip address after a reboot by setting network interface to eth0 instead of ens160


manifests/modules/packer/templates/vsphere/debian.rb.erb
```diff
@@ -25,7 +25,7 @@ end
  
 puts '- Re-obtaining DHCP lease...'
 
-<% if @operatingsystemmajrelease == '6' -%>
+<% if @operatingsystemmajrelease == '6' or @operatingsystemmajrelease == '8' -%>
```
gimmyxd: could write this as `if @operatingsystemmajrelease in ['6', '8']`

PR #455, opened 2019-05-17
(IMAGES-1114) Add packer support for Fedora 30

GabrielNagy:
Added Fedora 30 to puppetlabs-packer:
- installed `libxcrypt-compat` for libcrypt.so.1 required by Puppet-provided ruby
- disabled consistent network device renaming (keep eth0)

### DIFF
--- a/manifests/modules/packer/manifests/networking/params.pp
+++ b/manifests/modules/packer/manifests/networking/params.pp
@@ -40,7 +40,7 @@ class packer::networking::params {
           $udev_rule     = undef
           $udev_rule_gen = undef
         }
-        '29': {
+        '29', '30': {
           $interface_script = undef
           $udev_rule        = undef
           $udev_rule_gen    = undef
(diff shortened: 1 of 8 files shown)


GabrielNagy:
> @GabrielNagy The `30` sub-directory under `fedora` is also required to provide a `vars.json` and any other files associated with Fedora 30

The file is here: [templates/fedora/30/x86_64/vars.json](https://github.com/puppetlabs/puppetlabs-packer/pull/455/files#diff-6279cece0d2a38a4a83194668d04f4ee), if this is what you are referring to.

jcoconnor:
Indeed you are correct - not sure why I missed that - apologies.

templates/fedora/common/vars.json
```diff
@@ -1,6 +1,6 @@
 {
   "floppy_dirs"                          : "",
   "floppy_files"                         : "../../common/files/ks.cfg",
-  "boot_command"                         : "<tab> inst.text inst.ks=hd:fd0:/ks.cfg <enter>",
+  "boot_command"                         : "<tab> inst.text inst.ks=hd:fd0:/ks.cfg<enter>",
```
gimmyxd: This change is not needed, you could revert it

GabrielNagy: I changed that since the space is redundant but I can revert it if it's not in the scope of this PR.

PR #458, opened 2019-05-28
(maint) Use non-beta repos for rhel8

underscorgan:


manifests/modules/packer/manifests/vsphere/repos.pp
```diff
@@ -97,8 +97,7 @@
         # We don't have consistent mirror urls between RedHat versions:
         # TODO: RHEL 5 needs further refactoring
         $base_url = $facts['operatingsystemmajrelease'] ? {
-          # TODO: RHEL 8 is beta atm, we need to update with some local mirros when available
-          '8' => 'https://downloads.redhat.com/redhat/rhel/rhel-8-beta',
+          '8' => ""${repo_mirror}/rpm__remote_rhel-8",
```
donoghuc: Is there a duplicate `"` here?

underscorgan: gooood catch

manifests/modules/packer/manifests/repos.pp
```diff
@@ -25,24 +25,23 @@
...
         gpgcheck => '1',
         gpgkey   => 'file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-beta,file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release'
       }
 
-      yumrepo { 'rhel-8-for-x86_64-appstream-beta-rpms':
-        descr    => 'rhel-8-for-x86_64-appstream-beta-rpms',
-        baseurl  => "${base_url}/appstream/${::architecture}",
+      yumrepo { 'localmirror-appstream':
+        descr    => 'localmirror-appstream',
+        baseurl  => "${base_url}-appstream",
```
donoghuc: Out of curiosity what determines whether we use a `-` or a `/`? Artifactory `URL` endopoints? I'm looking at an old RHEL7 commit and they appear to be `/` for separating `baseurl`. https://github.com/puppetlabs/puppetlabs-packer/commit/049f3a7c24b0fbd81a6d6cdc1141135b1ef427e5

underscorgan: All the rhel repos that are mirrored from artifactory use the `-` due to how the repos are named and how that corresponds to file layout. Our old mirrors would mirror all the repos for an os version to a single directory, so each repo would be in a subdirectory

PR #463, opened 2019-08-30
(IMAGES-1137) Create machine image for winfips 2012r2

Dorin-Pleava:
Create machine image for winfips-2012r2-x64

### DIFF
--- a/templates/win/2012r2-fips/x86_64/files/platform-packages.ps1
+++ b/templates/win/2012r2-fips/x86_64/files/platform-packages.ps1
@@ -1,0 +1,17 @@
+
+. C:\Packer\Scripts\windows-env.ps1
+
+Write-Output "Running Win-2012r2-fips Package Customisation"
+
+Write-Output "Enabling FIPS mode"
+Set-ItemProperty -Path HKLM:\SYSTEM\CurrentControlSet\Control\Lsa\FipsAlgorithmPolicy -name Enabled -value 1
+
+if (-not (Test-Path "$PackerLogs\DesktopExperience.installed"))
+{
+  # Enable Desktop experience to get cleanmgr
+  Write-Output "Enable Desktop-Experience"
+  Add-WindowsFeature Desktop-Experience
+  Touch-File "$PackerLogs\DesktopExperience.installed" 
+}
+
+if (Test-PendingReboot) { Invoke-Reboot }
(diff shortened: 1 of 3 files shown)


templates/win/2012r2-fips/x86_64/files/platform-packages.ps1
```diff
@@ -0,0 +1,17 @@
+
+. C:\Packer\Scripts\windows-env.ps1
+
+Write-Output "Running Win-2012r2-fips Package Customisation"
+
+Write-Output "Enabling FIPS mode"
+Set-ItemProperty -Path HKLM:\SYSTEM\CurrentControlSet\Control\Lsa\FipsAlgorithmPolicy -name Enabled -value 1
```
jcoconnor: This is an acceptable way do enable FIPS, but it might be preferable to put it into the Puppet code.
(which would make an interesting case of how do you start putting options in that trigger different configs from the puppet code).

Dorin-Pleava: As the puppet code that is run is shared, and the windows 2012r2 and 2012r2 fips platforms use the same image, it would require some more development effort to make some triggers for different configs for puppet code.

Dorin-Pleava: Could not find a way to convert the winfips powershell script to puppet code as the puppet code is shared, and I couldn't find a way to differentiate between two windows images, because they are the same.

Dorin-Pleava:
Tested this PR and it successfully generated a windows 2012r2 with FIPS mode enabled.

templates/win/2012r2-fips/x86_64/files/platform-packages.ps1
```diff
@@ -0,0 +1,17 @@
...
+
+if (-not (Test-Path "$PackerLogs\DesktopExperience.installed"))
+{
+  # Enable Desktop experience to get cleanmgr
+  Write-Output "Enable Desktop-Experience"
+  Add-WindowsFeature Desktop-Experience
+  Touch-File "$PackerLogs\DesktopExperience.installed" 
+}
+
+if (Test-PendingReboot) { Invoke-Reboot }
```
jcoconnor: Minor nit here but it needs fixing - can you add a LF at the end.
(You may need to check your editor settings to do this automatically).

templates/win/2012r2-fips/x86_64/files/platform-packages.ps1
jcoconnor: That's fine for the moment @Dorin-Pleava as discussed - we can mark that for later.

PR #464, opened 2019-09-19
(maint) disable name-service/cache on Solaris 11.4

ciprianbadescu:
on Solaris 11.4, name-service/cache uses old nscd application that doesn't know how to handle DNAME record


gimmyxd:
@ciprianbadescu  can you add a `(maint)` or ticket number on the commit message?

PR #465, opened 2019-09-26
(QENG-7455) Mitigation for Firewall/Network issues

jcoconnor:
Perform some mitigation actions for the for QENG-7455 by setting the Network Interfaces to private on first (post-clone boot). #469 also updates the VMWare Tools drivers. Ultimately we should also update the Cygwin Stack, but this is deferred on request from @mihaibuzgau until the Lovejoy release is out.

Also:
- Change use the module ayohrling/local_security_policy to set the security policies in puppet instead of applying a policy file (IMAGES-845)
- Move the bootcfg commands into puppet code
- Eliminate the config-winsettings.ps1 provisioner as all this code is now done in puppet.
- Fixup beaker-hostgenerator string for fips build so that it runs acceptance suite.
- Add ```win-2019-dev``` and ```win-2019-ssh``` platforms to support onward development of the windows build and in particular separate it from Cygin.


jcoconnor:
Have tested this on Windows-10 and 2019 - all appears to work nicely except the bootlog - so looking at that.

jcoconnor:
I've tested this on Windows 10/2019 successfully:
1. Firewall is disabled:
```
11:03:52     vcenter-iso: Info: Applying configuration version '1569923848'
11:03:52     vcenter-iso: Notice: /Stage[main]/Windows_template::Firewall::Firewall/Windows_firewall_profile[public]/state: state changed 'on' to 'off'
11:03:52     vcenter-iso: Notice: /Stage[main]/Windows_template::Firewall::Firewall/Windows_firewall_profile[private]/state: state changed 'on' to 'off'
11:03:52     vcenter-iso: Notice: /Stage[main]/Windows_template::Firewall::Firewall/Windows_firewall_profile[domain]/state: state changed 'on' to 'off'
```
2. Bootcfg is run:
```
11:03:52     vcenter-iso: Notice: /Stage[main]/Windows_template::Bootcfg::Bootcfg/Exec[Enable Bootlog]/returns: The operation completed successfully.
11:03:52     vcenter-iso: 
11:03:52     vcenter-iso: Notice: /Stage[main]/Windows_template::Bootcfg::Bootcfg/Exec[Enable Bootlog]/returns: executed successfully
```
3. Manually verified autologon is ok (it's actually already set in the Unattend, but the puppet ensure is still useful).
4. Security Policies are run (only one change is actually needed).
```
11:04:31     vcenter-iso: Notice: /Stage[main]/Windows_template::Policies::Security_policies/Local_security_policy[Password must meet complexity requirements]/policy_value: policy_value changed '1' to '0'
```
5. Windows environment (and therefore autologon check) is performed after all sysinternals work is done:
```
11:03:52     vcenter-iso: Notice: /Stage[main]/Windows_template::Apps::Sysinternals_pkgs/Registry::Value[Sysinternals_EULA_PsSuspend]/Registry_value[HKLM\DEFUSER\Software\Sysinternals\PsSuspend\\EulaAccepted]/ensure: created
11:03:52     vcenter-iso: Notice: /Stage[main]/Windows_template::Apps::Sysinternals_pkgs/Registry::Value[Sysinternals_EULA_PsTools]/Registry_key[HKLM\DEFUSER\Software\Sysinternals\PsTools]/ensure: created
11:03:52     vcenter-iso: Notice: /Stage[main]/Windows_template::Apps::Sysinternals_pkgs/Registry::Value[Sysinternals_EULA_PsTools]/Registry_value[HKLM\DEFUSER\Software\Sysinternals\PsTools\\EulaAccepted]/ensure: created
11:03:52     vcenter-iso: Notice: /Stage[main]/Windows_template::Apps::Sysinternals/Windows_env[PATH=C:\Packer\SysInternals]/ensure: created
11:03:52     vcenter-iso: Notice: /Stage[main]/Windows_template::Apps::Gitforwin/Download_file[Git-2.20.1-64-bit.exe]/File[download-Git-2.20.1-64-bit.exe.ps1]/ensure: defined content as '{md5}4fae44cff9256fcac07f632166759dcb'
```
Going to do final check on this on all windows platforms.

jcoconnor:
The Firewall settings are failing on Windows French, so putting DNM on this until I resolve.

templates/win/common/scripts/common/puppet-configure.ps1
```diff
@@ -17,7 +17,7 @@ $PuppetCounterFile = "$PackerLogs\Puppet.Counter"
 
 if (-not (Test-Path "$PuppetCounterFile")) {
   New-Item -Path $PuppetCounterFile -Force
-  Set-Content -Path $PuppetCounterFile "0" -Force
+  Set-Content -Path $PuppetCounterFile "0"
```
joshcooper: This partially reverts the change in https://github.com/puppetlabs/puppetlabs-packer/pull/465/commits/aeb8491454133d101cd73eab56c48c427ea7b8fb. Based on the commit message "This change will be merged back into earlier commit once working." were you planning fixing up?

templates/win/2012r2-fips/x86_64/vars.json
```diff
@@ -1,6 +1,6 @@
 {
     "template_name"     : "win-2012r2-fips-x86_64",
-    "beakerhost"        : "windows2012r2-fips-64",
+    "beakerhost"        : "windows2012r2-64",
```
joshcooper: Did you intend to remove this commit?

joshcooper:
Generally 👍 just have a few questions though none are blocking:

* Maybe drop the firewall module since it doesn't support French?
* Should we switch beaker-hostgenerator windows fips string?
* For non-Windows platforms we allow incoming ssh connections and leave the firewall running. Then it's up to tests to handle the firewall, for example, in puppet we turn the firewall off on the master, https://github.com/puppetlabs/beaker-puppet/blob/188ac379a4d5c5c4e009dd6a4573a476c7388643/setup/common/025_StopFirewall.rb. I assume PE does the same. I can see wanting to keep the firewall running on Windows for consistency. In which case, it may be better to selectively allow incoming tcp connections on port 22 and likely 445 for windows smb. I'm not sure about bolt testing though, as they require winrm and winrm/https.

templates/win/2012r2-fips/x86_64/vars.json
jcoconnor: Going to keep this one in place until at least this image build.
We don't have the packaging provider up just yet that I can grab the right puppet-agent (fips) for the acceptance test.
So ticketing this for later treatment.

gimmyxd: packages for windowsfips will be available only for puppet-agent master. Not sure which stream is used by the windows acceptance step.

jcoconnor:
@joshcooper Have taken a look at your comments:
1. Think you are correct on the firewall, so going to go for more specific rules and make sure they are uniform across all 3 profiles (which is actually what the ticket really identifies as the problem).
2. Need to keep the firewall module in place until we sunset the Windows 2008/r2 builds - once that's done we can drop it in favour of powershell commands to reduce branching.
3. Need to hold onto the standard windwos2012r2 bhg definition for the moment - may ticket for some later non-urgent work.

Change coming up for review again tomorrow once I've soaked it thru the builds.

templates/win/common/scripts/common/puppet-configure.ps1
jcoconnor: Yeah - as mentioned in slack - was doing some quick testing - this has all been fixed up now and works across all the different windies versions.

jcoconnor:
This is ready to go - all images have been built and successfully deployed to pooler.

templates/win/2012r2-fips/x86_64/vars.json
ciprianbadescu: beaker hostname is: windowsfips-2012-x64: https://github.com/puppetlabs/beaker-hostgenerator/blob/master/lib/beaker-hostgenerator/data.rb#L1450

jcoconnor:
For clarity - the Beaker Hostname is used purely for the acceptance test for this pipeline only.
Strictly the value here is "incorrect" to ensure the acceptance test ran ok - it will be fixed properly once the fips packages are up and running.

PR #466, opened 2019-09-27
(IMAGES-1145) enable ipv6 on redhat 7 fips

ciprianbadescu:
If the interface cfg file contains "File Managed by Puppet", it will be configured as before, otherwise Network Manager default configuration will be enabled for it


jcoconnor:
Adding @sbeaulie to review as well.
This looks 👍 to me but don't know my linux interfacen as well as the windies ones.

PR #468, opened 2019-10-08
(MAINT) Adding a 7.2 image with tmpfs support

mikkergimenez:
This is my first time working in this repo.  I want to add a template to CI where tmpfs is enabled in centos 7.2.  This may become the default config, but we want to test this first, so here it is.  I've tested this manually by cloning a vm and running a command, now I want to get it into vmpooler.  The goal here is to reduce the I/O load in CI.  

Not sure if I should do this as 7.2-tmpfs or x86_64-tmpfs but I put it this way because the ks.cfg is in the common folder, and it's the ks.cfg config I need to fix.  

I looked at this documentation:
https://confluence.puppetlabs.com/display/SRE/Linux+Image+Packer+Generation

But did I miss anything further that would suggest how to test this?  I tried reverse engineering the Jenkins job but that was no fun.

### DIFF
--- a/templates/centos/7.2-tmpfs/x86_64/vars.json
+++ b/templates/centos/7.2-tmpfs/x86_64/vars.json
@@ -1,0 +1,14 @@
+{
+    "template_name"                         : "centos-7.2-tmpfs-x86_64",
+    "template_os"                           : "centos7_64Guest",
+    "beakerhost"                            : "centos7-64",
+    "version"                               : "0.0.1",
+    "iso_url"                               : "https://artifactory.delivery.puppetlabs.net/artifactory/generic/iso/CentOS-7-x86_64-DVD-1511.iso",
+    "iso_checksum"                          : "907e5755f824c5848b9c8efbb484f3cd945e93faa024bad6ba875226f9683b16",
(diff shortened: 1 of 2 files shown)


mikkergimenez:
That CI error seems suspicious.

jcoconnor:
@mikkergimenez This looks ok - I'll run it through test tomorrow.

templates/centos/7.2-tmpfs/x86_64/vars.json
```diff
@@ -0,0 +1,13 @@
+{
+    "template_name"                         : "centos-7.2-x86_64",
```
jcoconnor: template name should be ```centos-7.2-tmpfs-x86_64``` - this is the actual template name that appears up on vcenter.

jcoconnor: Sorted in latest push

jcoconnor:
I've pushed some changes up with allow the OS to boot and install - but its not establishing connection back to packer once the installation is complete.

Going to discuss with @ciprianbadescu tomorrow who is authoring a similar PR #467

PR #469, opened 2019-10-10
(MAINT) Fixes and Updates to Windows Imaging

jcoconnor:
This is a set of small non-conflicting changes to Puppet-Labs packer in preparation for a general windows Image refresh. It includes
1. Adding locale to the build data (originally associated with QENG-7455 but useful in its own right).
2. Fix minor warning in the puppet counter file loop
3. Use JSON lookup instead of Env-Var data for the puppet run (Server Type lookup)
4. Update Version number for new refresh
5. Remove win-2012r2-fr-x86_64 and win-2016-fr-x86_64 as 2019 now replaces this
6. Update Windows 10 to Build 1903 and add Build 1909 as next
7. Update VMWare tools to assist with QENG-7455


jcoconnor:
These commits have been tested successfully and build and run acceptance suite without any issues.

PR #471, opened 2019-10-11
Add entropy generating tools to fips image

shrug:
Our test pipelines for FIPS builds are rapidly exhausting the available pool of entropy.

this adds two tools to fips images help increase the available entropy pool - rng-tools and haveged

This is "fake" entropy and should not be relied on for production, but should be fine for our testing purposes

### DIFF
--- a/templates/redhat/7.2-fips/common/files/ks.cfg
+++ b/templates/redhat/7.2-fips/common/files/ks.cfg
@@ -22,6 +22,9 @@ auth  --useshadow  --enablemd5
 firstboot --disabled
 reboot --eject
 
+repo --name=epel-release --baseurl=https://artifactory.delivery.puppetlabs.net/artifactory/rpm__remote_epel/7/x86_64/
+
+
 %packages --ignoremissing
 @core
 bzip2
@@ -34,6 +37,7 @@ net-tools
 patch
 perl
 rng-tools
+haveged
 curl
 wget
 nfs-utils
--- a/templates/redhat/7.2-fips/common/files/ks.cfg
+++ b/templates/redhat/7.2-fips/common/files/ks.cfg
@@ -33,6 +33,7 @@ make
 net-tools
 patch
 perl
+rng-tools
 curl
 wget
 nfs-utils
--- a/templates/redhat/7.2-fips/x86_64/vars.json
+++ b/templates/redhat/7.2-fips/x86_64/vars.json
@@ -2,7 +2,7 @@
     "template_name"                         : "redhat-fips-7.2-x86_64",
     "template_os"                           : "rhel7_64Guest",
     "beakerhost"                            : "redhatfips7-64",
-    "version"                               : "0.0.8",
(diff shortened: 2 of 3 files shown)


shrug:
This has been built and should be deploying to the pooler now for testing

shrug:
Ping @gimmyxd or @ciprianbadescu (or team) - Does this look ok? I guess the alternative is to do the installation and configuration in the puppet code instead. I did run a build but the templates are not deployed.

shrug:
We probably should deploy this to unblock fips testing, but I'm not sure I know how we're doing that these days and it's probably better that I don't try late on a Friday :)

templates/redhat/7.2-fips/common/files/ks.cfg
```diff
@@ -22,6 +22,9 @@ auth  --useshadow  --enablemd5
 firstboot --disabled
 reboot --eject
 
+repo --name=epel-release --baseurl=https://artifactory.delivery.puppetlabs.net/artifactory/rpm__remote_epel/7/x86_64/
+
+
```
ciprianbadescu: keep only one empty line

templates/redhat/7.2-fips/common/files/ks.cfg
```diff
@@ -22,6 +22,9 @@ auth  --useshadow  --enablemd5
 firstboot --disabled
 reboot --eject
 
+repo --name=epel-release --baseurl=https://artifactory.delivery.puppetlabs.net/artifactory/rpm__remote_epel/7/x86_64/
```
ciprianbadescu: I prefer repo change will go to: https://github.com/puppetlabs/puppetlabs-packer/blob/master/manifests/modules/packer/manifests/repos.pp, as all repo are set in one plce

shrug: Fair enough. It's in the kickstart because I also added the haveged package there (which comes from EPEL) but I can also move that into the manifests

shrug: We really only need the repo and those packages for FIPS rhel builds right now, and I don't see a really good way to distinguish FIPS builds in the puppet modules - any suggestions?

shrug:
Do you mind picking this up and running with it? FIPS test pipelines are blocked as the initial installation fails due to running out of entropy and /dev/random blocking

justinstoller:
fwiw, I created this ticket for us to track this work: https://tickets.puppetlabs.com/browse/IMAGES-1164

templates/redhat/7.2-fips/common/files/ks.cfg
ciprianbadescu: Needs to updates puppet/facter version in vars.json to get a newer fact that implements fips_enabled fact.
Also I see we don't have yet proper place to install apps on non-Windows.

I propose to merge this PR like it is and modify the ticket scope to handle refactoring and handling apps install/configuration in manifests.

PR #475, opened 2020-01-14
(IMAGES-1168) Add macOS 10.15 machine image

GabrielNagy:
Refer to the updated README for peculiarities regarding this image. Still have to test it in the pooler, but it builds on vCenter (with some manual interaction).


jcoconnor:
This looks ok to me but would prefer someone more familiar with MacOS builds to hit the 👍

templates/common/vmware.vcenter.json
```diff
@@ -35,12 +35,14 @@
       "vmware_base_required_modules"                 : "saz-ssh puppetlabs-stdlib",
 
       "vmware_base_vmx_data_memsize"                 : "4096",
+      "vmware_base_vmx_data_usb"                     : "FALSE",
```
GabrielNagy: I think this isn't used anywhere

PR #477, opened 2020-01-20
(IMAGES-1174) January 2020 Windows Patch Tuesday Fixes

jcoconnor:
A set of fixes that were needed to deliver the January 2020 Patch Tuesday Windows Refresh:
1. Fix Puppet Agent link to use Puppet 6
2. Disable WSUS for Windows 10/2016/2019
3. Add Servicing Stack Updates (SSU) for certain builds to allow update process to run


jcoconnor:
Fyi - there may be a few other fixen here so going to roll them up in one PR, hence the DNM.

PR #478, opened 2020-01-30
(maint) Upgrade Packer to 1.5.1, fix template errors

highb:
This commit upgrades the version of packer installed by the `Makefile`
to 1.5.1. After doing that upgrade, I found that several templates now
failed the tests because `_comment` fields were provided as arrays,
which is apparently unsupported now. I fixed up the arrays to be regular
strings and all of the templates appear to pass the tests now.

### DIFF
--- a/templates/win/2012/x86_64/x86_64.vmware.slipstream.json
+++ b/templates/win/2012/x86_64/x86_64.vmware.slipstream.json
@@ -13,9 +13,7 @@
 
   "description": "Customised Win-2012 build to prepare slipstream ISO",
 
-  "_comment": [
-      "The boot_command is hacky because the UEFI boot file used requires the 'Press any key' to be done"
-  ],
+  "_comment": "The boot_command is hacky because the UEFI boot file used requires the 'Press any key' to be done",
   "builders": [
     {
       "name": "{{user `template_name`}}-{{user `provisioner`}}-{{user `template_config`}}",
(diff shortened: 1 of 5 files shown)


GabrielNagy:
I wanted to build the macOS 10.15 template with packer 1.5.1, but the vsphere plugin wasn't compiled against that version so I had to downgrade back to 1.4 in order for it to work. With the vsphere builder now being integrated into packer I think it's a change we'd want to have. However a packer version including the vsphere builder wasn't released yet (it'll probably get into 1.5.2).

jcoconnor:
@GabrielNagy Wasn't aware that the vsphere builder was being integrated into packer - is that going to replace the jetbrains plugin completely ?

I've been following the packer releases and had intended to look at 1.5.1

GabrielNagy:
Yep https://github.com/hashicorp/packer/pull/8480

jcoconnor:
🏆 Super to see that happening - probably will be 1.6 - which should be due in February - they release almost monthly.

jcoconnor:
So - can we merge this now or do we need to wait until 1.5.1 is installed ?

GabrielNagy:
I don't think it matters really, from what I understand this is just executed as setup for tests that validate the template syntax, right @highb? Having comments as strings works on both 1.4 and 1.5, so I'm good with this whenever.

highb:
Yeah, this is only used by the `Makefile` for testing the templates (as well as the `Rakefile` I'm writing to simplify some of the image build process)

PR #482, opened 2020-03-27
(MAINT) Use Packer vsphere-iso/clone plugins

jcoconnor:
Packer has not integrated the jetbrains vsphere-* plugins into its core
so switch over all vsphere builds to use this in favour of our local
versioned plugin.

This means the plugins can be removed from the builders.
One less thing to maintain - Yayyyyy

### DIFF
--- a/templates/common/vmware.vcenter.json
+++ b/templates/common/vmware.vcenter.json
@@ -71,7 +71,7 @@
 
  "builders": [
     {
-      "type"                                         : "vsphere-23-iso",
+      "type"                                         : "vsphere-iso",
 
       "communicator"                                 : "{{user `communicator_protocol`}}",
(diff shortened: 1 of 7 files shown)


jcoconnor:
Adding DNM - just need to check this builds before we say GO GO GO

highb:
Once this is in, we should remember to remove the line about downloading the plugin from the docs https://confluence.puppetlabs.com/display/AGENT/Adding+new+vmpooler+image

highb:
🤞 Hope these build with no issues.

jcoconnor:
So it doesn't build because Packer on the build machines is 1.3.2

I will discuss with @puppetlabs/night-s-watch on Monday before making any changes.

PR #484, opened 2020-04-02
(PE-28666) Add centos-8-pdbpreload-x86_64

nmburgan:
## (DIO-409) Add project_root to custom script path
Because project_root seems not to be available within vars.json of a template, this sets the custom_provisioning_script to add the project_root to the beginning in the top level vmware.vcenter.json, so the user just has to specify 'scripts/myscript.sh' for custom_provisioning_script.

## (PE-28666) Add centos-8-pdbpreload-x86_64
This adds a new centos 8 template that includes setting up PE with the latest master build, and importing the SLV soak database. This will be used for HA testing.

## (PE-28666) Manually partition centos 8
Because autopart only allocates 50GB max to / and gives the rest to /home, and we need more than this for pdbpreload nodes (and the base centos 8 image so replicas can be provisioned), this changes to manually setting up the root, swap and boot partitions.

### DIFF
--- a/scripts/pdbpreload.sh
+++ b/scripts/pdbpreload.sh
@@ -1,0 +1,230 @@
+#!/bin/bash
+
+# These should be passed in via custom_provisioning_env
+#BRANCH='master'
+#VERSION='latest'
+#PLATFORM='el-8-x86_64'
+
+PE_TARBALL_URL="https://artifactory.delivery.puppetlabs.net/artifactory/generic_enterprise__local/${BRANCH}/ci-ready"
+EXT='.tar'
+BACKUP_DB_URL='http://slv-performance-results.s3-website-us-west-2.amazonaws.com/releases/Kearney/SLV-415/kearney_soak/master'
+BACKUP_TAR='db_backup.2019.05.29.tar.gz'
+TMP_DIR='/tmp'
+PUPPET_BIN='/opt/puppetlabs/bin'
+RUN_TUNE=1
+
+set -e
+
+# This isn't really used since we use %{::trusted.certname} below,
+# but sometimes we end up getting a spicy-proton name here due to
+# some sort of IP reuse thing, and it could cause weirdness.
+HOSTNAME=$(hostname -f)
+echo "### HOSTNAME = ${HOSTNAME} ###"
+
+df -h
+lsblk
+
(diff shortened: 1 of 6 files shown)


scripts/pdbpreload.sh
```diff
@@ -0,0 +1,230 @@
...
+echo '### Running puppet one more time just to be safe ###'
+${PUPPET_BIN}/puppet agent -t
+
+# The refresh_master_hostname plan requires user_data.conf to be present
+echo '### Running puppet infra recover_configuration to generate user_data.conf ###'
+${PUPPET_BIN}/puppet-infrastructure recover_configuration
+
+echo '### Creating refresh_hostname script ###'
+echo BOLT_DISABLE_ANALYTICS=true /opt/puppetlabs/installer/bin/bolt --boltdir=/opt/puppetlabs/installer/share/Boltdir plan run enterprise_tasks::testing::refresh_master_hostname > /root/refresh_hostname.sh
+chmod +x /root/refresh_hostname.sh
```
highb, line 228: One little tweak that could be made here is to pop a systemv script in here that enables itself on bootup and then disables the script? Maybe that's more of a pain than just kicking it off with Beaker, though.

nmburgan: Yeah, that would be nice.  Problem is, this takes several minutes to run, and requires PE services to be able to start.  I'm not sure where we'd be at when this runs on boot.  Could give it a shot.

highb: I don't think it should block this PR. Just a thought that occurred to me.

highb:
Testing this out now.

highb:
@nmburgan Were you able to get this to build from your local/platform9 system? It looks like the host is failing to get it's kickstart file for me.
Screen Shot 2020-04-06 at 3 50 32 PM
Screen Shot 2020-04-06 at 3 53 29 PM

nmburgan:
No, this was entirely tested on vcenter.

nmburgan:
FWIW, the current image used for centos-8-pdbpreload-x86_64 was built using this PR.  Not quite sure why kickstart is unhappy locally.  That's strange.

highb:
Might just be something weird with my platform9 box? 🤷‍♂

PR #485, opened 2020-04-08
Migrate centos to vcenter

sbeaulie:
adding required iso_name and setting valid template_os for vcenter 6.7
as per https://vdc-download.vmware.com/vmwb-repository/dcr-public/da47f910-60ac-438b-8b9b-6122f4d14524/16b7274a-bf8b-4b4c-a05e-746f2aa93c8c/doc/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html


sbeaulie:
This passes `packer validate` so that all the required variables are present, but a lot of them still need to confirm the boot_command works.

Goes with https://github.com/puppetlabs/ci-job-configs/pull/6955

PR #488, opened 2020-05-27
(PE-29139) Add packer template work for Centos7/ Irving mono preinstall (WIP)

actowery:
(WIP)
This PR creates new packer template data for the creation of a preinstalled centos 7/2019.1.15 mono node to be used in CI.

PR #489, opened 2020-05-28
(PE-29139) Add packer template work for Centos7/ Irving mono preinstall

actowery:
This PR creates new packer template data for the creation of a preinstalled centos 7/2018.1.15 mono node to be used in CI.

replaces #488

### DIFF
--- a/scripts/mono-preload.sh
+++ b/scripts/mono-preload.sh
@@ -47,127 +47,125 @@ ${PUPPET_BIN}/puppet agent -t
 # The refresh_master_hostname plan requires user_data.conf to be present
 echo '### Running puppet infra recover_configuration to generate user_data.conf ###'
 ${PUPPET_BIN}/puppet-infrastructure recover_configuration
+
+# Stop all services and disable puppet here instead of at the beginning of the refresh 
+# hostname script. This saves time in CI
 puppet resource service puppet ensure=stopped enable=false
+puppet agent --disable
+puppet resource service pe-nginx ensure=stopped
+puppet resource service pe-console-services ensure=stopped
+puppet resource service pe-puppetserver ensure=stopped
+puppet resource service pe-bolt-server ensure=stopped
+puppet resource service pe-ace-server ensure=stopped
+puppet resource service pe-orchestration-services ensure=stopped
+puppet resource service pe-puppetdb ensure=stopped
+puppet resource service pe-postgresql ensure=stopped
+
 
 echo '### Creating refresh_hostname script ###'
 
 set -e
-cat <<'EOF' > /opt/puppetlabs/installer/share/Boltdir/site-modules/enterprise_tasks/plans/testing/refresh_master_hostname.pp
-# Note: This plan requires user_data.conf to be present. We can't run
-# it from inside this plan since the plan is meant to be run AFTER
-# the hostname has already changed and therefore
-# puppet infra recover_configuration will not work correctly.
-plan enterprise_tasks::testing::refresh_master_hostname(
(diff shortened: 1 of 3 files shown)


scripts/mono-preload.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+# These should be passed in via custom_provisioning_env
+#BRANCH='master'
+#VERSION='latest'
+#PLATFORM='el-8-x86_64'
+
+PE_TARBALL_URL="https://artifactory.delivery.puppetlabs.net/artifactory/generic_enterprise__local/${BRANCH}/ci-ready"
```
nmburgan, line 8: This won't work for releases, only dev versions.  Releases are at `https://artifactory.delivery.puppetlabs.net/artifactory/enterprise/archives/releases/${VERSION}`

scripts/mono-preload.sh
```diff
@@ -0,0 +1,65 @@
...
+
+set -e
+
+# This isn't really used since we use %{::trusted.certname} below,
+# but sometimes we end up getting a spicy-proton name here due to
+# some sort of IP reuse thing, and it could cause weirdness.
+HOSTNAME=$(hostname -f)
+echo "### HOSTNAME = ${HOSTNAME} ###"
+
+df -h
```
nmburgan, line 21: You can get rid of this line and the next.  It was for debugging when we were adding extra disks.

scripts/mono-preload.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# This isn't really used since we use %{::trusted.certname} below,
+# but sometimes we end up getting a spicy-proton name here due to
+# some sort of IP reuse thing, and it could cause weirdness.
+HOSTNAME=$(hostname -f)
+echo "### HOSTNAME = ${HOSTNAME} ###"
+
+
+# Read LATEST file to grab the latest build from the directory
+if [ -z "${PE_VERSION}" ]; then
```
nmburgan, line 23: There's a mismatch between `VERSION` and `PE_VERSION` here.  You are specifying `VERSION` as an env var in vars.json, but you probably want `PE_VERSION`.

If this file is only ever meant to be used for creating pdbpreload images of releases (based on using the release tarball URL above), you can get rid of checking `LATEST` and continue to use `VERSION` as the env var passed in.

templates/centos/7.0-mono-2018.1.15-preload/x86_64/vars.json
```diff
@@ -0,0 +1,15 @@
...
+    "beakerhost"                            : "centos7-64",
+    "version"                               : "0.0.1",
+    "iso_url"                               : "http://osmirror.delivery.puppetlabs.net/iso/CentOS-7.0-1406-x86_64-DVD.iso",
+    "iso_checksum"                          : "713ea7847adcdd1700e92429f212721a",
+    "iso_checksum_type"                     : "md5",
+    "vmware_vsphere_nocm_vmx_data_memsize"  : "4096",
+    "vmware_vsphere_nocm_vmx_data_numvcpus" : "2",
+    "puppet_aio"                            : "http://yum.puppetlabs.com/el/7/PC1/x86_64/puppet-agent-1.8.3-1.el7.x86_64.rpm",
+    "iso_name"                              : "CentOS-7.0-1406-x86_64-DVD.iso",
+    "custom_provisioning_env"               : "BRANCH=2018.1,VERSION=2018.1.15,PLATFORM=el-7-x86_64",
```
nmburgan, line 13: Don't need `BRANCH` since we aren't using a dev build and are only using `VERSION`

PR #497, opened 2020-11-13
(IMAGES-1231) November 2020 Windows Image Refresh

jcoconnor:
Generate the Windows November 2020 Patch Tuesday Image Refresh.

This picks up serious CVE-2020-3941 as well as updates to:
1. Powershell 7.1.0
2. NotePad++ 7.9.1
3. Google chrome 86.0.4240.198
4. Git for Windows 2.29.2.2
5. vmware tools 11.2.0 (via boot process)

Windows 10 is also updated to:
1. win-10-ent: Build 2004
2. win-10-next: Build 2009 (20H2)

Fixed issue with pswindows update testing - seems to be a bug in This
since Build 2004 where the get-wuhistory hangs - adding last 2 to the
command appears to be a vaid workaround.

Google appears to install in the Program Files directory (and no longer
in the x86 one) so adjust in scripting.

Filter our Packer Environment Files on the Forcefully delete so as to
avoid throwing (albeit non-fatal) errors in the cleanup scripts at the
end.

Create PowerCLI script to fixup images for new netapp storage pending
a fix in platform-ci-utils and ci-jobs.

Enter these updates in the command record script.

### DIFF
--- a/templates/win/10-ent/i386/vars.json
+++ b/templates/win/10-ent/i386/vars.json
@@ -8,15 +8,15 @@
     "windows_version"   : "Windows-10",
     "image_name"        : "Windows 10 Enterprise",
     "product_key"       : "NPPR9-FWDCX-D2C8J-H872K-2YT43",
-    "iso_name"          : "en_windows_10_business_editions_version_1909_updated_july_2020_x86_dvd_9588cbe4.iso",
+    "iso_name"          : "en_windows_10_business_editions_version_2004_updated_july_2020_x86_dvd_de266803.iso",
     "iso_checksum_type" : "md5",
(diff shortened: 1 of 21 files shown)


mattkirby (approved):

sbeaulie (approved):
nicely done
The fixup script is useful, as we may be able to use it again for moving stuff around in the future. But wouldn't it have been easier to fix the platforms-util code instead? I can't imagine that the datastore is more than a couple lines of hardcoded values to change?

jcoconnor:
💯 on fixing it in platform-ci-utils - which I'm looking at now (that's the real reason I'm holding on the ci-job-configs).
But it took ages to build the images and was easier to just fix them up before deployment.

heathseals (approved):

PR #500, opened 2021-01-22
(RE-13438) Remove old GPG key & obsolete PC1 references

e-gris:
As part of updating the new release signing key, I found some code
here that also referenced the long-removed PC1 repo.

Deleted it.

### DIFF
--- a/manifests/modules/packer/manifests/puppet.pp
+++ b/manifests/modules/packer/manifests/puppet.pp
@@ -3,59 +3,8 @@
 # A define that manages puppet
 #
 class packer::puppet {
-
-  case $facts['osfamily'] {
-    debian: {
-      include apt
-
-      apt::source { 'puppetlabs-pc1':
-        location   => 'http://apt.puppetlabs.com',
-        repos      => 'PC1',
-        key        => '7F438280EF8D349F',
-        key_server => 'pgp.mit.edu',
-      }
-
-      package { 'puppet-agent':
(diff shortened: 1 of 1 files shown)


manifests/modules/packer/manifests/puppet.pp
```diff
@@ -3,59 +3,8 @@
...
-        ensure  => present,
-        require => Yumrepo[ 'puppetlabs-pc1' ],
-      }
-    }
-
-    default: {
-      fail( "Unsupported platform: ${::osfamily}/${::operatingsystem}" )
-    }
-  }
-
```
underscorgan, line 55: I don't know how this code is used but I think this class may need updates to add more current repositories rather than deleting everything. Not sure who maintains this though :\

e-gris: That's definitely the question. 

My (possibly incorrect) assertion is that since none of this actually exists anymore and its lack of function seems to be bothering nobody, its existence is questionable.

GabrielNagy: You're right @e-gris, this doesn't appear to be used anywhere so I agree we should remove it to avoid confusion.

GabrielNagy: Actually the entire repo can benefit from a thorough clean-up, but this is a good start.

GabrielNagy (approved):

PR #503, opened 2021-02-09
(PE-31039) Fix refresh_hostname script for 2019.8.4

nmburgan:
In 2019.8.4, we set up the PE Infrastructure Agent node group to manage master_uris, pcp_broker_list, and server_list. We need to change the values in here to the new hostname as part of the refresh_hostname script.  Post-2019.8.4, we can add this to the refresh_master_hostname plan so we don't need this patch in the bash file we lay down on the image. But for now, this patch will fix the refresh_hostname.sh script on the 2019.8.4-based image.


nmburgan:
This apparently runs into problems when checking out a node and running it (but works if you just run it twice).  But this is currently working just fine in CI.  Something to do with resolving localhost.localdomain I guess.  I think for now, we can just merge this since we're using the image based on this, and I'm going to move this fix into the plan so we don't need it in the future.

PR #505, opened 2021-05-03
(REPLATS-168) Add a k8s restart script for kurl image

jpartlow:
When a kurl vm is checked out, there's a multi step process to get
containerd and the k8s services running again. It becomes more
complicated when the host ip has changed and k8s services need to be
reconfigured for the new ip.

In order to make it simple for anyone manually checking out a kurl vm to
reset it to a working state, this patch pulls down the restart-k8s.sh
script from the kurl_test repo into /root so it's available to be run
when a vm is checked out.

There doesn't seem to be a straight forward mechanism for including
arbitrary files with puppetlabs-packer without modifying the
templates/common/vmware.vcenter.json to include an additional file
provisioner. But that would affect all the images using that template.

### DIFF
--- a/scripts/kurl.sh
+++ b/scripts/kurl.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 # These should be passed in via custom_provisioning_env
-#APP='cd4pe'
-#CHANNEL='beta'
+#APP='puppet-application-manager'
+#CHANNEL='standalone-beta'
 
 set -e
 
--- a/scripts/kurl.sh
+++ b/scripts/kurl.sh
@@ -6,6 +6,22 @@
 
 set -e
 
+#################################################################################
+# Provide a script to get kubernetes back into a working state after vm checkout.
+sudo yum install -y git
+/usr/bin/git clone git@github.com:puppetlabs/kurl_test
+sudo cp kurl_test/tasks/restart_k8s.sh /root
+sudo chmod 750 /root/restart_k8s.sh
+rm -rf kurl_test
+
+sudo cat << EOF >> /etc/motd
+
+After initial checkout, run /root/restart_k8s.sh to get the Kubernetes environment restarted.
+
+EOF
+
+#############################
+# Install Kubernetes via Kurl
 curl -sSLO https://k8s.kurl.sh/bundle/${APP}-${CHANNEL}.tar.gz
 tar xzf ${APP}-${CHANNEL}.tar.gz
 rm ${APP}-${CHANNEL}.tar.gz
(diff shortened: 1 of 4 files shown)


jpartlow:
Not sure yet if the curl will work here.

scripts/kurl.sh
```diff
@@ -1,11 +1,25 @@
...
+
+After initial checkout, run /root/restart_k8s.sh to get the Kubernetes environment restarted.
+
+EOF
+
+#############################
+# Install Kubernetes via Kurl
 curl -sSLO https://k8s.kurl.sh/bundle/${APP}-${CHANNEL}.tar.gz
 tar xzf ${APP}-${CHANNEL}.tar.gz
 rm ${APP}-${CHANNEL}.tar.gz
```
MikaelSmith, line 25: The patch is redundant now because it doesn't come with prometheus.

scripts/kurl.sh
```diff
@@ -1,11 +1,25 @@
 #!/bin/bash
 
 # These should be passed in via custom_provisioning_env
-#APP='cd4pe'
-#CHANNEL='beta'
+#APP='puppet-application-manager'
+#CHANNEL='standalone-beta'
```
MikaelSmith, line 5: Tempted to switch to the stable `standalone` channel. We probably don't need to update this frequently, and it makes more sense to pull down a beta bundle as a one-off if we need to test out changes.

scripts/kurl.sh
jpartlow: Yes, that's true; will we want it for the ha nodes though? Also it provides a template for any other config changes we want to make to kurl in the future.

jpartlow: Removed; can always refer to the git history.

scripts/kurl.sh
jpartlow: Switched to standalone.

jpartlow:
Hmm, the two test builds I tried failed for unrelated reasons. [Ubuntu 20.04](https://cinext-jenkinsmaster-sre-prod-1.delivery.puppetlabs.net/view/imaging/view/Ubuntu/job/imaging_ubuntu-20.04_x86_64_vmware_vcenter_packer_build_packer/) for package dependencies and [Ubuntu 18.04](https://cinext-jenkinsmaster-sre-prod-1.delivery.puppetlabs.net/view/imaging/view/Ubuntu/job/imaging_ubuntu-18.04_x86_64_vmware_vcenter_packer_build_packer/25) because of issues running platform-ci-utils...so maybe I'll stop poking around :)

MikaelSmith (approved):

PR #508, opened 2021-05-18
(IMAGES-1259) Fedora 34 fixes

Dorin-Pleava:
Disabled systemd-resolved and restored /etc/resolv.conf as
normal file.

### DIFF
--- a/manifests/modules/packer/templates/vsphere/redhat.rb.erb
+++ b/manifests/modules/packer/templates/vsphere/redhat.rb.erb
@@ -85,7 +85,7 @@ Kernel.system('/sbin/service NetworkManager restart')
 
 puts '- Cleaning up...'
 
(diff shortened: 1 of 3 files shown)


ciprianbadescu (approved):

manifests/modules/packer/templates/vsphere/redhat.rb.erb
```diff
@@ -85,7 +85,7 @@ Kernel.system('/sbin/service NetworkManager restart')
 
 puts '- Cleaning up...'
 
-<% if ['28', '29', '30', '31'].include? @operatingsystemrelease -%>
+<% if ['28', '29', '30', '31', '34'].include? @operatingsystemrelease -%>
```
GabrielNagy, line 88: should have probably made this apply to all fedoras >=28... or just all fedoras since we won't build versions older than 28 anymore

PR #509, opened 2021-06-25
(REPLATS-169) Bump centos-8.3-kurl image version to 0.1.1

jpartlow:
...to bring in Weave 2.8.1. We've been seeing transient issues with
weave networkpolicy rules not being up to date after a restarted vm and
want to see if pulling in the newer puppet-application-manager builds
that include a newer weave-net improve the problem.


genebean (approved):
LGTM

PR #510, opened 2021-07-15
Replats 348 add ha kurl image

jpartlow:
(REPLATS-348) Add an unformatted partition to 8.3-kurl-ha for Ceph

The Ceph persistent volume manager for k8s needs to work with an
unformatted device. To start with, I've just carved out 16384MiB from the
71680MB sda volume. This is considerable smaller than the recommended
50GB + app req that we recommend, but may suffice for CI.

---

    (REPLATS-348) Only preload kurl in the 8.3-kurl-ha image

Restarting k8s was challenging for a standalone primary. It's likely to
be worse for an ha primary with ceph/rook, and then we'd need to worry
about joining other primaries to the cluster for ha.

Instead, we're just going to pre-download the kurl tarball, which should
be the bulk of the time, and then install fresh on checked out vms.

Curl's puppet-application-manager-stable instead of standalone, and
updates the kurl.sh to skip installation and managing a restart script
(which we aren't copying into the image) if we are not installing.

### DIFF
--- a/scripts/kurl.sh
+++ b/scripts/kurl.sh
@@ -3,26 +3,33 @@
 # These should be passed in via custom_provisioning_env
 #APP='puppet-application-manager'
 #CHANNEL='standalone'
+#INSTALL=true (default, set false to to skip installation)
+INSTALL="${INSTALL:-true}"
 
 set -e
 
 #################################################################################
 # Provide a script to get kubernetes back into a working state after vm checkout.
 
-sudo mv /tmp/restart_k8s.sh /root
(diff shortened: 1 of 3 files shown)


jpartlow:
I'll test this out building to vmpooler-dev.

jpartlow:
Work on this is temporarily blocked waiting for a kurl issue to be resolved: https://tickets.puppetlabs.com/browse/REPLATS-407

yachub:
Closing since this project will be archived shortly.

PR #513, opened 2021-08-12
(REPLATS-360) Bump centos-8.3-kurl image version to 0.1.4

jpartlow:
...and switch channel from standalone to standalone-beta. There are
changes from the past month that we want to test before the next
release.


jpartlow:
@MikaelSmith switching the kurl vm to standalone-beta channel.

MikaelSmith (approved):

PR #514, opened 2021-08-13
(maint) Fix missing images for airgap tests

MikaelSmith:
Stops Kubelet before shutdown. The packer build fills the disk with 0s to compress the image, which otherwise causes Kubelet to start erasing unused images that we still need.


jpartlow (approved):

jpartlow:
I'll get an instance of this into vmpooler-dev for us to look at.

MikaelSmith:
This prevents evicting images.

PR #517, opened 2021-08-18
(maint) Drain pods before imaging

MikaelSmith:
We're seeing transient issues with pods starting up after reinitializing via kubeadm. Drain all pods before we stop kubelet and create the image to ensure pods have a chance to safely shutdown.

Also prevents kubelet and containerd from starting on startup as the first thing we do in the restart script is to stop them. This avoids having to cleanup pods again.

### DIFF
--- a/scripts/kurl.sh
+++ b/scripts/kurl.sh
@@ -24,7 +24,6 @@ curl -sSLO https://k8s.kurl.sh/bundle/${APP}-${CHANNEL}.tar.gz
 tar xzf ${APP}-${CHANNEL}.tar.gz
 rm ${APP}-${CHANNEL}.tar.gz
 
-dnf -y install bash-completion
 cat install.sh | sudo bash -s airgap preserve-selinux-config
 
 # Stop Kubelet before shutdown. The packer build fills the disk with 0s to compress the image,
--- a/scripts/kurl.sh
+++ b/scripts/kurl.sh
@@ -26,6 +26,29 @@ rm ${APP}-${CHANNEL}.tar.gz
 
 cat install.sh | sudo bash -s airgap preserve-selinux-config
 
-# Stop Kubelet before shutdown. The packer build fills the disk with 0s to compress the image,
-# which otherwise causes Kubelet to start erasing unused images that we still need.
+# Stop pods and Kubelet before shutdown. The packer build fills the disk with 0s to compress the
+# image, which otherwise causes Kubelet to start erasing unused images that we still need. Other
+# pods may also be paused in an unexpected state if left running.
+echo " * Draining node"
+kubectl drain localhost.localdomain --delete-local-data --ignore-daemonsets
+
+echo " * Stopping kubelet"
 systemctl stop kubelet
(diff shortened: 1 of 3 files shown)


MikaelSmith:
Creating a dev build with https://jenkins-sre.delivery.puppetlabs.net/view/imaging/view/Cent%20OS/job/imaging_centos-8.3-kurl-beta_x86_64_vmware_vcenter_packer_build_packer/46/console. Going to poke at it to see if the image seems significantly different than our 0.1.6 build.

MikaelSmith:
I'm not sure this fixes anything, but it would probably make me more confident in the build. We can use this to pull in new updates to restart_k8s.sh.

MikaelSmith:
Hmm, this seems to reliably break envoy for some reason, in ways that look similar to https://github.com/projectcontour/contour/issues/3775.

scripts/kurl.sh
```diff
@@ -27,6 +27,29 @@ rm ${APP}-${CHANNEL}.tar.gz
...
 
-# Stop Kubelet before shutdown. The packer build fills the disk with 0s to compress the image,
-# which otherwise causes Kubelet to start erasing unused images that we still need.
+# Stop pods and Kubelet before shutdown. The packer build fills the disk with 0s to compress the
+# image, which otherwise causes Kubelet to start erasing unused images that we still need. Other
+# pods may also be paused in an unexpected state if left running.
+echo " * Draining node"
+kubectl drain localhost.localdomain --delete-local-data --ignore-daemonsets
+
+echo " * Stopping kubelet"
```
MikaelSmith, line 36: Pulled directly from https://github.com/puppetlabs/kurl_test/blob/main/tasks/restart_k8s.sh#L47-L64, but also disable services so they don't start when the instance starts up. Restart script should handle starting them when it's ready.

/cc @jpartlow

MikaelSmith:
Seems like https://jenkins-sre.delivery.puppetlabs.net/view/imaging/view/Cent%20OS/job/imaging_centos-8.3-kurl-beta_x86_64_vmware_vcenter_packer_build_packer/48/console took a long time to download the airgap image, but finally produced a build.

jpartlow:
That build had the:

```
==> vcenter-iso: dd: error writing '/EMPTY': No space left on device
```

message, does it still have the images?

MikaelSmith:
Yes, that happens every time because it tells `dd` to fill the disk.

MikaelSmith:
This image seems to work, and resets pretty quickly. However I suspect some of our problems happen when vmpooler is moderately to heavily loaded, which doesn't happen in the dev-pool.

jpartlow:
Oh, right; the problem wasn't the disk being zero'd, it was kubelet panicking about the disk being zero'd.

jpartlow (approved):

MikaelSmith:
Made a few more small changes, going to do one more rebuild to check.

jpartlow:
firmer-dime, which is an image with the previous version (before imaging pipeline's #49 run) has a hung envoy container.

PR #522, opened 2021-08-30
(DIO-2470) Replace vsphere authorized keys

mattkirby:
This change updates the authorized keys for vsphere images. Without this change we use an old authorized keys file.


yachub:
Closing since this project will be archived shortly.

PR #528, opened 2021-11-17
(REPLATS-514) Update kurl image to Redhat 8.4

jpartlow:
Build a new kurl-beta image from redhat 8.4, and attempt to have it take
care of k8s restart during it's checkout process by handling it with a
systemd service.

* Switch from centos-8.3 to redhat-8.4 since centos is no longer
supported
* Moves restart_k8s.sh from
https://github.com/puppetlabs/kurl_test/blob/04c3eec9d7d9924923f11af27821260afb26f28d/tasks/restart_k8s.sh
to templates/redhat/8.4-kurl-beta/common/files
* Adds a File provisioner to the templates/common/vmware.vcenter.json to
copy this file from the project to the image
* Adds a systemd service during the custom kurl script to run this
restart_k8s.sh on startup
* Get latest pam standalone-beta (for free when we build this image)

The centos-8.3-kurl-beta image definition is removed from puppetlabs-packer as part of this, but it's still available from ABS, atm.

---

Also adds a workaround for the [PT-582](https://tickets.puppetlabs.com/browse/PT-582) dnf kurl.local module bug ￼…

Kurl's local repositories are configured as dnf modules. This leaves
behind a @modulefailsafe definition for kurl.local tied to the last
local repository kurl installed from. This in turn locks packages that
are in that repository to the repository versions and causes future
upgrade problems.

Calling dnf module reset clears out that stored module metadata,
allowing packages to update normally.


MikaelSmith (approved):

jpartlow:
Made some changes to the service dependencies. I'm going to rebuild the image, get it into staging and test that out again.

scripts/kurl.sh
```diff
@@ -67,3 +53,27 @@ fi
...
+chown root:root /usr/sbin/restart_k8s.sh
+chmod 755 /usr/sbin/restart_k8s.sh
+cat > "/usr/lib/systemd/system/k8s-restart.service" <<SERVICE
+[Unit]
+Description=Kubernetes network restart for abs checkout
+After=network-online.target rc-local.service
+Wants=network-online.target rc-local.service
+
+[Service]
+Type=oneshot
```
MikaelSmith, line 68: Should we also use `Restart=on-failure`?

jpartlow: I'm inclined to start with the simpler case of works/doesn't work and only switch to restarts if we find we can't fix a base problem and restarting resolves it.

MikaelSmith (approved):

templates/common/vmware.vcenter.json
```diff
@@ -229,6 +229,12 @@
                                                      ]
     },
 
+    {
+      "type"                                         : "file",
+      "sources"                                      : "{{user `custom_files_to_upload`}}",
+      "destination"                                  : "/tmp/"
+    },
+
```
ciprianbadescu, line 237: This looks to break all images builds not defining `custom_files_to_upload`. Is there other way to implement this?

ciprianbadescu: @mhashizume , @joshcooper fyi

MikaelSmith: If this is standard gotemplate, we could wrap the whole block in something like
```
{{if (user `custom_files_to_upload`)}}
...
{{end}}
```

MikaelSmith: Looks like that probably wouldn't work. Maybe provide a default value that refers to an empty file.

MikaelSmith: https://github.com/puppetlabs/puppetlabs-packer/pull/531 is one option to test out.

jpartlow: When I pulled the custom_files_to_upload back in from the 3318e6a revert, I failed to include the /dev/null default var, I'll place that back in.

jpartlow: I opened #532, which also reverses #530 because we need to pass more than one file; let me know if there's something else about that commit that's required. Sorry for the problems Ciprian!

jpartlow: (As far as *other* other approaches, if we're keeping puppetlabs-packer as the mechanism for building the images in the new pipeline world, we could redesign the custom script/files approach used for specialized images like the kurl image as puppet classes, and instead provide a param for including additional classes in the puppet builders. But I don't want to spend time on that until we know more about the new image build process and decide we want to refactor here.)

ciprianbadescu: thanks for fixing this so quick

PR #532, opened 2021-12-02
(REPLATS-514) Fix custom files to upload default

jpartlow:
When I pulled the file provisioner back in from 3318e6a, I forgot to
include the default value for custom_files_to_upload. Consequently other
images using vmware.vcenter.json as their base template would fail,
since they had no value set here. Readding as /dev/null so that the
builder is effectively a noop unless an image specifies something to
copy.

---

@mhashizume this also reverts 5b90b4 because when I'm using custom_files_to_upload, I'm passing an array, so I believe sources is the required parameter. If you made that change because you were trying to correct my mistake with the default, then we're probably okay? But if something else was broken, please let me know. Thanks!

### DIFF
--- a/templates/common/vmware.vcenter.json
+++ b/templates/common/vmware.vcenter.json
@@ -232,7 +232,7 @@
 
     {
       "type"                                         : "file",
-      "source"                                       : "{{user `custom_files_to_upload`}}",
+      "sources"                                      : "{{user `custom_files_to_upload`}}",
       "destination"                                  : "/tmp/"
     },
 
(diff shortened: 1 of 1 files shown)


templates/common/vmware.vcenter.json
```diff
@@ -231,7 +232,7 @@
 
     {
       "type"                                         : "file",
-      "source"                                       : "{{user `custom_files_to_upload`}}",
+      "sources"                                      : "{{user `custom_files_to_upload`}}",
```
MikaelSmith, line 235: It's surprising that this works with a string rather than array of strings. But presumably it's what you used when you were testing builds.

jpartlow: Yes, it is a bit odd. I recall trying to pass it as an array at one point, and that failing, but I don't remember the details.

MikaelSmith (approved):

templates/common/vmware.vcenter.json
jpartlow: Here it is, I think: https://www.packer.io/docs/templates/legacy_json_templates/user-variables#using-array-values

PR #533, opened 2021-12-06
(PT-582) Remove dnf module workaround from kurl.sh script

jpartlow:
This was working around a bug in the local dnf repositories used by Kurl
when building the redhat-8.4-kurl-beta image.

As of https://kurl.sh/release-notes/v2021.12.01-0, kurl includes this
workaround itself. Consequently, the kurl.sh script was failing because
dnf would exit non-zero when we tried to `dnf module reset kurl.local`
because Kurl's install.sh script had already removed it...

### DIFF
--- a/scripts/kurl.sh
+++ b/scripts/kurl.sh
@@ -15,10 +15,6 @@ rm "${APP}-${CHANNEL}.tar.gz"
 
 cat install.sh | sudo bash -s airgap preserve-selinux-config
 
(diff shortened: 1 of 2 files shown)


jpartlow:
Building here: https://jenkins-sre.delivery.puppetlabs.net/view/imaging/view/Redhat/job/imaging_redhat-8.4-kurl-beta_x86_64_vmware_vcenter_packer_build_packer/3

MikaelSmith (approved):

PR #536, opened 2021-12-23
(maint) Removes legacy OSes

mhashizume:
This removes support for several operating systems that are both end-of-life and are no longer supported by Puppet.

Additionally, this includes a small README update that replaces instances of "Puppet Labs" with "Puppet." (Drop the "labs," it's cleaner.)


jpartlow (approved):
:+1: Looks like a good cleanup.

PR #537, opened 2022-02-09
Replats 591 bump pam on kurl beta

jpartlow:


jpartlow:
This was built successfully here: https://jenkins-sre.delivery.puppetlabs.net/view/imaging/view/Redhat/job/imaging_redhat-8.4-kurl-beta_x86_64_vmware_vcenter_packer_build_packer/7

And vmpooler has been updated with that vm.

jpartlow:
There are problems with prometheus, will be adding another commit to exclude it.

jpartlow:
Ok, was able to install using that vm from abs-stage. The vmpoolerProd version bump is puppetlabs/infracore-k8s-configs#528

jpartlow:
Image 0.3.5 is running in ci.

PR #547, opened 2022-04-27
Focal updates

underscorgan:


manifests/modules/packer/manifests/vsphere/repos.pp
```diff
@@ -75,7 +75,7 @@
       }
 
       # IMAGES-1217, for some reason the rebuild of ubuntu 18.04 was missing bionic-updates
-      if $facts['operatingsystem'] == 'Ubuntu' and $facts['operatingsystemrelease'] == '18.04' {
+      if $facts['operatingsystem'] == 'Ubuntu' and $facts['operatingsystemrelease'] in ['18.04', '20.04'] {
```
nmburgan, line 78: Would it make sense to do operatingsystemmajrelease >= 18 instead?  Looking ahead to the day when we add 22 and 24.

underscorgan: I'm hoping by the time we do that we're fully on the new gcp system 🙂

joshcooper (approved):

manifests/modules/packer/manifests/vsphere/repos.pp
nmburgan: Good point!

nmburgan (approved):

